PR #142, opened 2014-04-16
Css refractor rework

g-knee:
screen shot 2014-04-16 at 5 16 19 am
screen shot 2014-04-16 at 5 16 55 am
screen shot 2014-04-16 at 5 17 44 am
screen shot 2014-04-16 at 5 18 08 am

New Features: redesigned filter bar, map now displays full width on mobile/small enough browsers

Minor changes: current location button in pale green on home screen, changes to search button to make it look nicer

Paradigm shifts: something to think about for future designers is that to allow the map to display in full, I had to change container to being full width, and isolating everything that is currently displayed as being padded inwards and wrapping it in a class "constrained". So, if you're making a new page in haml, be sure that if you want your text to be displayed in a container of the proper width, put it within a div with the "constrained" class.

Bugs caused by me: none yet apparent


kiesia:
I'm not sure if this work is related, but I just noticed #197

PR #192, opened 2015-01-29
Feature/display current year in copyleft

kiesia:
It's the little things that annoy.. Keep the year up to date on the copyleft notice.


tkwidmer:
+1 thanks!

PR #193, opened 2015-01-29
Initialize dotenv vars correctly in debian, simplify conf

### LINKED ISSUE #170: Devise.secret_key was not set
Whenever I try to run the app (ex. with pow) I get this error:

RuntimeError: Devise.secret_key was not set. Please add the following to your Devise initializer:

  config.secret_key = '-----'

Please ensure you restarted your application after installing Devise or setting the key.

I know we set the key in devise.rb, so I'm surprised to get this error. I get the same error when I change the key environment variable to the new key given in the error message and restart the app.

kiesia:
This commit fixes #170 on Debian at least.


kiesia:
@tkwidmer would be nice to get this checked on OSX an merged in, as the build is currently broken on (at least) debian.

tkwidmer:
wont this override the production env variable set in heroku?

tkwidmer:
Thats why we had test and development secret keys set. and didnt set just a general env variable.

kiesia:
@tkwidmer the dotenv gem is only bundled in the test and development environments, so it shouldn't be an issue on Heroku.

tkwidmer:
k, wanted to make sure

PR #198, opened 2015-01-30
Users can propose updates to restrooms (versioning)

kiesia:
**This is not ready to merge, I'm making a pull request now so I have a place to ask questions when they come up.**

 * Add versioning to restrooms with Papertrail.
 * Create admin interface for dealing with multiple restroom versions and submission approval.


app/controllers/restrooms_controller.rb
```diff
@@ -62,7 +57,25 @@ def destroy
     end
   end
 
+  def upvote
```
tkwidmer: The issue with moving these to custom actions is then we are not following strict REST / CRUD. Which has advantages and disadvantages. As a note. The original code I wrote had these as custom actions, but it was suggested by many that we stick to a strict CRUD since these are just attributes on the table there is no need to have custom actions. 

However, I realize with papertrail it may not be possible to accept updating these columns without making them custom actions. In this case. Is it possible to use just one endpoint /vote or something? This endpoint could taken in a parameter for upvote / downvote ?

kiesia:
@tkwidmer the strict CRUD argument doesn't work when we introduce complexity. 

I refactored this because the voting should not be versioned, but a normal update should be (not only versioned, but we should roll back the version in the update method too). At this point the method has multiple responsibilities and should be refactored into multiple methods. I also felt it was simpler to write two simple vote methods than create a longish method that had to deal with working out which attribute to update, if any.

I haven't added the code to roll back the updated restroom, so this is still a bit unclear. Does this make sense?

tkwidmer:
I'm ok with the separate endpoints (it's how I originally built the controller). But might want to get input from others.

tkwidmer:
an interesting relevant discussion: http://discuss.emberjs.com/t/custom-rest-actions-non-crud-actions/5718/10

I like the suggestion of if we do use a custom action having it just be one endpoint which routes to different actions depending on if its a post or a delete. ie: 

```
POST   /api/v1/places/1/vote
DELETE /api/v1/places/1/vote
```

kiesia:
:+1:

tkwidmer:
Although as a coworker just pointed out that is less clear to clients who might consume the API than /upvote /downvote

-- But it does take advantage of the HTTP methods available to us, which is a more proper API design. Essentially treating it as a sub-resource // sudo-resource. That can be created (upvoted) and deleted. (downvoted)

tkwidmer:
Since this has stalled, i'm going to close this for now.

PR #202, opened 2015-01-31
Remove restroom destroy and edit actions

kiesia:
Cause we don't want someone to delete the database for the lulz..


kiesia:
This has now been deployed.

PR #205, opened 2015-02-05
reload postgres config after change during vagrant setup

mi-wood:
@tkwidmer I assume you were having the same issue setting up vagrant. This should fix it :)


mi-wood:
Oh, was also running out of memory. I don't think the app is going to run without 1GB unfortunately :(

tkwidmer:
looks good to me.

PR #206, opened 2015-02-09
Integrate Piwik

kiesia:
Add Piwik support. Analytics javascript will only display on production. I left the config in ```config/piwik.yml```. Not sure if we want that in an ENV variable, there is nothing sensitive there.


tkwidmer:
I dont see the need for an env var. Looks good to me.

tkwidmer:
Something about Piwik or our setup of it is blocking the application from functioning on the new Heroku Cedar-14.

PR #208, opened 2015-02-09
Don't strip dashes on restroom addresses

### LINKED ISSUE #207: Titelized restroom address strips dashes
Dashes are very useful with addresses! See: http://www.refugerestrooms.org/restrooms/4638 It should read Admiralstr. 1-2, but looks like 1 2.

kiesia:
Fixes #207


tkwidmer:
Seems fine to me. Not sure if there was a particular reason it was being titlecized.

PR #210, opened 2015-02-09
Downgrade grape + grape-swagger to fix url issues

### LINKED ISSUE #203: Cannot access API at http://www.refugerestrooms.org/api/docs/
When I head to the API link, I get this error:

*Unable to read api 'restrooms' from path http://www.refugerestrooms.org/api/swagger_doc.json/restrooms.json (server returned undefined)*

kiesia:
Fixes #203. I'm not super familiar with the API or grape, so please double check that all is still working.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,8 +24,8 @@ gem 'country_select'
 gem 'mail_form'
 gem 'simple_form', '~> 3.0.2'
 
-gem 'grape'
-gem 'grape-swagger'
+gem 'grape', '0.7.0'
+gem 'grape-swagger', '~> 0.7.2'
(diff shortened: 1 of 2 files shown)


tkwidmer:
Seems fine to me. As long as its showing up on the docs page it should be fine.

hkellaway:
Im unfortunately still getting the error

cllns:
It looks like it's only been merged into `develop`, not `master`.

In general, changes to `develop` should be reflected on `http://refugestaging.herokuapp.com` but it seems like `/api/` is not a part of staging.

tkwidmer:
I can do a release today to master.

PR #212, opened 2015-02-10
Titleize top city name

cllns:
Right now it says "San francisco, CA" instead of "San Francisco, CA". 

This fixes that!


tkwidmer:
great, thanks @cllns !

PR #219, opened 2015-02-23
vagrant fixes

tkwidmer:
add apt update, remove versioning from postgres packages, run postgres commands as postgres user


tkwidmer:
this work is from of @mi-wood !

PR #230, opened 2015-03-22
Mi wood/bug/226 persistent filter

mi-wood:
@tkwidmer @kevjames3 WIP
I don't know this javascript thing, so somebody tell me if I'm doing everything wrong. But here's what's in this PR:
- Move filters to the header and match css to search bar (does not display on welcome page)
- Use query string to toggle search for ada/unisex
- Add js function to change query string on click

Still needs tests. There are a few more comments inline 

The screenshot below looks like there's a bit too much padding between the search bar and the filters.
screen shot 2015-03-22 at 3 09 30 am


app/assets/javascripts/views/restrooms/restrooms.js
```diff
@@ -0,0 +1,49 @@
+/*
+Place all the behaviors and hooks related to the matching controller here.
+All this logic will automatically be available in application.js.
+*/
+$(function(){
```
mi-wood: Is there something more rails or a coffeescript querystring class to replace this with?

app/views/layouts/_filters.html.haml
```diff
@@ -0,0 +1,24 @@
+/ Filters for ada, unisex and map toggle
+.row
+  .col-xs-12.headroom
+    #filters.clearfix
+      .btn-group{:data => {:toggle => "buttons"}}
+        - if params[:ada].present? && params[:ada] == 'true'
```
mi-wood: I know there's a way to combine the indented lines in haml (or maybe not?), but I got pretty sick of looking up haml docs.

app/controllers/restrooms_controller.rb
```diff
@@ -54,7 +54,17 @@ def update
 
 private
   def list_restrooms
-    @restrooms = Restroom.all.page(params[:page])
+    params[:ada] ||= false
+    params[:unisex] ||= false
+    if params[:unisex] == 'true' && params[:ada] == 'true'
+      @restrooms = Restroom.where('accessible AND unisex').page(params[:page])
```
mi-wood: unfortunately, you can't wildcard a boolean lookup (I tried :/) so this code is a bit repetitive

mi-wood:
screen shot 2015-03-22 at 3 19 55 am
screen shot 2015-03-22 at 3 20 08 am

mi-wood:
Also, I notice that if you're in mapview and try to change the filters, it will return to list view :/

mi-wood:
Weird alignment with the accessibility logos on that last screenshot

mi-wood:
Also, TODO might be to implement AJAX for this, but I have was running into a few issues. There's an example with the paginator gem here: https://github.com/amatsuda/kaminari_example/tree/ajax/app

mi-wood:
ok, started to refactor index to coffeescript, since I'm going to be touching the toggle map feature for this anyways. this is not a working branch ATM. Screenshot below shows that the map/list can be toggled now and will auto-toggle the opposite side, so that only one is active. I'm not doing anything to make it not possible to change that feature and allow a map and list at the same time.
screen shot 2015-03-22 at 10 28 03 pm

mi-wood:
If anyone has a sec to look at this, I got stuck with how the map is displaying. The button functionality works, but the map appears with the thumbtack icons and no map behind it! Haven't been able to figure out what's going on

PR #236, opened 2015-05-06
Reduce size of city-photo.jpg background picture, and add retina versions

cllns:
The current background image is massive. This PR reduces the size of the original background image, while adding larger versions for retina screens.
For me, on coffee shop wifi, the current image takes 4-7 seconds to load. I imagine it's much worse for people on phones.

New files are:
- `city-photo.jpg` is **321K** and 1600 x 804 pixels
- `city-photo-2x.jpg` is **1MB** and 3200 x 1608 pixels
- `city-photo-3x.jpg` is **4.2MB** and 7516 x 3777 pixels (the current background image)

One thing that's a potential problem is that phones with retina screens will likely load `city-photo-2x.jpg`, even though they'd be fine with `city-photo.jpg`. I'm not sure how to solve that without adding a heavier weight solution, like http://responsiveimages.org/.


tkwidmer:
gtg ty <3 @cllns

PR #240, opened 2015-06-10
Noticed the lower-case H in GitHub

brntbeer:
Hope this doesn't come across as an offensive PR! Just noticed common spelling error in the GitHub name

Let me know if these aren't the correct places to make these edits!


mi-wood:
Thanks!

PR #245, opened 2015-10-07
Add specific permitted params for restrooms

cllns:
Found this vulnerability with `brakeman`. 

Instead of calling `permit!` we should be using strong params to allow only the params we want.


mi-wood:
lgtm :dancer:

tkwidmer:
gtg

Issue #241, opened 2015-06-19
Add blockList Bathrooms list


tkwidmer:
This allows people to report bathrooms at which the have been harassed by employees at in the past.

strand:
So, don't read the comments, but you might consider [blocklist/safelist](http://english.stackexchange.com/questions/51088/alternative-term-to-blacklist-and-whitelist).

tkwidmer:
Thats much better verbage.

apfritts:
+1

azelma:
I'll work on this.

ShayShayBeekay:
As this is attached to another issue, anyone know if it's safe to close?

strand:
@ShayShayBeekay Thanks for pinging on this,

DeeDeeG:
For the historical record: 
- `edit_id` field for each revision
- Users can propose changes
- db has `approved` boolean for each revision

All introduced in: #530

PR #246, opened 2015-10-15
Add blocked flag to restrooms

valeriecodes:
This is one possible approach to https://github.com/RefugeRestrooms/refugerestrooms/issues/241

It adds a flag on restrooms for blocked, and a dropdown on the restroom form for users to add the blocked flag.
screen shot 2015-10-14 at 9 42 04 pm
Blocked restrooms have a "ban" icon on the show page.
screen shot 2015-10-14 at 9 44 11 pm

But this raises a couple questions:
- How to message "blocking," do you want to have a page that explains "blockable" conduct?
- How to deal with these restrooms in API queries, etc. since these are essentially anti-recommendations, we don't people to accidentally go to them when looking for a safe space.

Let me know your thoughts.

### DIFF
--- a/app/assets/stylesheets/restrooms.css.scss
+++ b/app/assets/stylesheets/restrooms.css.scss
@@ -70,7 +70,7 @@
         max-width: 100%;
     }
 
-    // ADA / Gender Neutral Filters
+    // ADA / Gender Neutral / Blocked Filters
     .itemExtraInfo {
         position: absolute;
         top: 20px;
@@ -83,6 +83,9 @@
     .ADARestroom {
         width: 44px;
         height: 47px;
+    }
+    .blockedRestroom {
+      color: red;
     }
     .unisexRestroom img, .ADARestroom img {
         width: 100%;
--- a/app/models/restroom.rb
+++ b/app/models/restroom.rb
@@ -41,6 +41,8 @@ class Restroom < ActiveRecord::Base
 
   scope :accessible, -> { where(accessible: true) }
   scope :unisex, -> { where(unisex: true) }
+  scope :blocked, -> { where(blocked: true) }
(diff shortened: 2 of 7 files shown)


tkwidmer:
I think how we handle this in the API will be important. This solution has a little bit more dangerous in terms of that. I know it saves us having to have a separate db table, controller, etc. But I'd almost prefer that it was a separate blocked_restrooms table. In terms of usability - i'd rather that people always be able to trust that hitting the restrooms endpoint will always produce a list of safe restrooms and not have to filter it for blocked / not blocked restrooms.

I'd like to hear what @mi-wood @strand @cllns and other have to say though.

azelma:
That seems totally fair and reasonable @tkwidmer. I'll wait for others to weigh in, but happy to implement it that way.

mi-wood:
My initial thoughts are that blocked restrooms should never appear to a user. There should be two ways to get them:
1. admin interface
2. explicitly setting a blocked=true flag (or separate table) on the api

We might not even want '2.' to be query-able though. My only thought is that when someone goes to add a restrooms, there should be a way to inform them that this is already blocked.

tkwidmer:
@mi-wood I think that kind of functionality would be useful, but probably al little complicated due to user error on addresses / names. Its going to be hard to match probably.

mi-wood:
@tkwidmer would lat/long not work?

tkwidmer:
I don't know margin of error between calls on geocoder. Like does an address always return the same lat long exactly? Or might it be a point off?

mi-wood:
@valeriecodes @tkwidmer Is this something we could close for now and maybe revisit? Unless someone is willing to pick it up atm, which would be rad.

Issue #233, opened 2015-03-28
Add if a restroom has a baby changing station


tkwidmer:
Got this email in through the contact us link

> Hi! I want to thank you so much for what you do and for offering the use of this website and I note I don't require this website (I'm cisgender) but am incredibly grateful this is here. You should be proud, so proud for giving safety for people who just need to go to the bathroom. Thank you so much--you are awesome!!!
> 
> Now to business: I would like to see this website integrate into it whether the restroom has a baby-changing station or not. This was called to my attention when I saw a Change.org petition to a certain company by Ashton Kutcher, who is a dad, and he was requesting that the men's restrooms at the company have a baby changing station for parenting. I then realized that they discriminated against him for being a dad who wanted to take care of his child. So then I realized that this website didn't have a baby-changing indicator for its restrooms on the site. I would greatly appreciate the addition of this feature to the editing section of the page to both keep "bathroom owners" accountable and to give a sense of comfort/less anxiety for trans_, genderqueer, and intersex parents. I am proud of this website for being disability-friendly, too! That is very smart and accommodating. I love you all to death figuratively for having this website. I am incredibly grateful. The good reputation of this website will only get better if baby-changing stations are added. Not only would this be trans_ (umbrella) friendly and disability-friendly, the page would be "parent-friendly" and by extension, "baby-friendly." This would be great and I highly recommend it.

azelma:
I'll take this.

tkwidmer:
Closed by #248

PR #248, opened 2015-10-18
Changing tables

azelma:
Add field to indicate whether a restroom has a changing table, per: https://github.com/RefugeRestrooms/refugerestrooms/issues/233

I don't know what icon should be used for this. The best I could come up with was the fa-child one, but I think there's probably better out there, open to suggestions.

Screenshots:
screen shot 2015-10-18 at 4 09 01 pm
screen shot 2015-10-18 at 4 08 42 pm

@tkwidmer

### DIFF
--- a/app/assets/javascripts/views/restrooms/restrooms.js.coffee
+++ b/app/assets/javascripts/views/restrooms/restrooms.js.coffee
@@ -3,15 +3,20 @@
 # You can use CoffeeScript in this file: http://coffeescript.org/
 
 $ ->
-  $('#ada_filter').click -> 
+  $('#ada_filter').click ->
     if $(this).hasClass("active")
       $('.listItem.not_accessible').show()
-    else 
+    else
       $('.listItem.not_accessible').hide()
-    
-  $('#unisex_filter').click -> 
+
+  $('#unisex_filter').click ->
     if $(this).hasClass("active")
       $('.listItem.not_unisex').show()
-    else 
+    else
       $('.listItem.not_unisex').hide()
 
+  $('#changing_table_filter').click ->
+    if $(this).hasClass("active")
+      $('.listItem.no_changing_table').show()
+    else
+      $('.listItem.no_changing_table').hide()
--- a/app/views/restrooms/_restroom.html.haml
+++ b/app/views/restrooms/_restroom.html.haml
@@ -1,4 +1,9 @@
-%div{:class => "listItem #{restroom.unisex? ? '' : 'not_unisex'} #{restroom.accessible? ? '' : 'not_accessible'} #{restroom.changing_table? ? '' : 'no_changing_table'}", "data-id" => restroom.id}
+- unisex_class = restroom.unisex? '' : 'not_unisex'
+- accessible_class = restroom.accessible? '' : 'not_accessible'
+- changing_table_class = restroom.accessible? '' : 'no_changing_table'
+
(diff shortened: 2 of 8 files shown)


app/assets/javascripts/views/restrooms/restrooms.js.coffee
```diff
@@ -3,15 +3,20 @@
...
-  $('#unisex_filter').click -> 
+
+  $('#unisex_filter').click ->
     if $(this).hasClass("active")
       $('.listItem.not_unisex').show()
-    else 
+    else
       $('.listItem.not_unisex').hide()
 
+  $('#changing_table_filter').click ->
```
mi-wood: It's possible that we'll want to start pulling out all these boolean icons to enums rather than listing all of them here. Doesn't need to be part of this though, just a follow up. By no means is three icons unwieldy :P

mi-wood: Also, there may be a more 'rails' way to do it. Enums is just the first thing that pops into my head.

app/views/restrooms/_restroom.html.haml
```diff
@@ -1,4 +1,4 @@
-%div{:class => "listItem #{restroom.unisex? ? '' : 'not_unisex'} #{restroom.accessible? ? '' : 'not_accessible'}", "data-id" => restroom.id}
+%div{:class => "listItem #{restroom.unisex? ? '' : 'not_unisex'} #{restroom.accessible? ? '' : 'not_accessible'} #{restroom.changing_table? ? '' : 'no_changing_table'}", "data-id" => restroom.id}
```
mi-wood: I'm not 100% sure it will work, but maybe break this into multiple lines. It seems haml might use `,` for that. haml has weird rules so ¯\_(ツ)_/¯

mi-wood:
A few minor comments. I don't have objections to the icon chosen, although maybe others have ideas

app/views/restrooms/_restroom.html.haml
tkwidmer: This could be done with a little bit of refactoring too.

```haml
- unisex_class = restroom.unisex? '' : 'not_unisex'
- accessible_class = restroom.accessible? '' : 'not_accessible'
- changing_table_class = restroom.accessible? '' : 'no_changing_table'

%div{ class: "listItem #{unisex_class} #{accessible_class} #{changing_table_class}",
          "data-id": restroom.id}
```

tkwidmer:
Other than that, I'm fine with this. =)

azelma:
Did that HAML refactor @tkwidmer @mi-wood

app/assets/javascripts/views/restrooms/restrooms.js.coffee
tkwidmer: I don't really think there is a rails convention that would help here. These technically are three different filters. There may be a gem that will help with this, but since we are doing this filtering in javascript -- I don't know if it would really help without a refactor here.  I'm fine with this for now.

tkwidmer:
gtg, gratz on your contribution! @azelma thank you!

app/assets/javascripts/views/restrooms/restrooms.js.coffee
cllns: I too don't think it's necessary at this point but I think we'd want to use a bitfield, not an enum. Here's a popular one that's named after a silly looking dog https://github.com/pboling/flag_shih_tzu :)

Issue #167, opened 2014-10-07
Rename WelcomeController to StaticController


tkwidmer:
The welcome controller has in practice been our 'pages other than restrooms' controller. Welcome really isn't a great name for it. We should name it something else. Suggestions?

azelma:
This looks like it's been stagnant for awhile, I'm going to take a crack at it.

tkwidmer:
fixed by #250

PR #250, opened 2015-10-28
Switch to high_voltage gem for static pages

azelma:
Addresses https://github.com/RefugeRestrooms/refugerestrooms/issues/167


tkwidmer:
i think this looks good.

tkwidmer:
Not sure why I never merged this. But I did now.

PR #253, opened 2015-12-19
I add a javascript validation to contact form

geobde:


app/views/layouts/application.html.haml
```diff
@@ -13,5 +13,7 @@
         = yield
         = render 'layouts/footer'
 
+    %javascript_include_tag 'validate', 'data-turbolinks-track' => true
```
tkwidmer: You should just include this into the asset pipeline rather than a include tag here.

app/assets/javascripts/validate.js
```diff
@@ -0,0 +1,50 @@
+//= require jquery
+//= require jquery_ujs
+//= require turbolinks 
+//= require_tree .
+
+var userName = document.getElementById('username');                            // Declare and set elements to variables.
+var userEmail = document.getElementById('email');
+var userMsg = document.getElementById('message');
+var valMessageName = document.getElementById('message-name');
```
tkwidmer: why not just have one div, and append each error message? or better yet, append the error message below the input?

tkwidmer:
you do have a failing build as well. https://travis-ci.org/RefugeRestrooms/refugerestrooms/builds/97880292

mi-wood:
@tkwidmer I'd like to have something like this, but since larger framework/refactor work is being talked about, I'm wondering if we close this for now?

PR #262, opened 2016-04-27
Also displays distance to restroom in kilometers alongside miles

hsson:
For all of us using the metric system this update also displays the distance to restrooms in kilometers :smile:

### DIFF
--- a/app/views/restrooms/_restroom.html.haml
+++ b/app/views/restrooms/_restroom.html.haml
@@ -24,7 +24,9 @@
     .itemDistance
       - if restroom.respond_to?(:distance)
         = restroom.distance.round(2)
-        miles
+        miles /
+        = (restroom.distance * 1.609344).round(2)
+        kilometers
     - if restroom.unisex?
       .unisexRestroom
         %i.fa.fa-transgender-alt.fa-3x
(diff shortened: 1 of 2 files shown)


app/views/restrooms/_restroom.html.haml
```diff
@@ -24,7 +24,9 @@
     .itemDistance
       - if restroom.respond_to?(:distance)
         = restroom.distance.round(2)
-        miles
+        miles /
+        = (restroom.distance * 1.609344).round(2)
```
tkwidmer: Can we put this in a helper somewhere? And define the 1.609344 inside the helper as a constant. Rather than just having a hardcoded amount here?

http://mixandgo.com/blog/the-beginner-s-guide-to-rails-helpers has an example of what i'm talking about. It would be something like this:

file_name: `app/helpers/restrooms_helper.rb`

```ruby
module RestroomHelper
  KILO_DISTANCE_CONVERSION = 1.609344
  def miles_to_kilometers(miles)
    miles * KILO_DISTANCE_CONVERSION
  end
end
```

then you should be able to use it here just by calling
`miles_to_kilometers(restroom.distance).round(2)`

tkwidmer:
otherwise this is a great idea!

hsson:
@tkwidmer Alright, I'll add it to a helper and update the pull request :+1:

hsson:
I refactored the conversion to a helper as suggested :smile:

tkwidmer:
looks ok to me, ty!

Issue #264, opened 2016-06-02
Location Services Disabled; Tell them they need to enable.


cs09g:
os: window 7
web browser: latest chrome

when click search button with no query, refreshing animation starts with no reaction.
at least it should show what do i do, and stop the animation because it does not search anything.

Hkly:
What's the expected behavior when searching without a query? The pull request referencing this https://github.com/RefugeRestrooms/refugerestrooms/pull/269 seems to imply that it searches the current location if no query is entered, but when I try that I'm just left spinning.

cs09g:
It does show the message that a query should be entered, but spinning icon doesn't stop.

Hkly:
Ah so it looks like you have location services blocked or disabled on your browser. When location services is enabled, searching with no query does a search on your current location.

A fix for this would maybe be adding a message letting users know they need to enable location services. And stop the animation spinning too :P

cs09g:
@Hkly Right! I guessed so. That fix you just mentioned would be great.

DeeDeeG:
I believe this is fixed by #359.

PR #269, opened 2016-09-04
Potential UX fix to issue #264.

CullenIO:
Adds a message to tell the user it is searching by location.

### DIFF
--- a/app/assets/javascripts/views/restrooms/search.coffee
+++ b/app/assets/javascripts/views/restrooms/search.coffee
@@ -5,6 +5,7 @@
 
 class Refuge.Restrooms.Search
   searchDefaultText: "1 Embarcadero Center, San Francisco, CA"
+  searchCurrentLocationText: "Searching by current location..."
 
   constructor: (form) ->
     # Define elements that will be in use in this class.
@@ -44,6 +45,7 @@ class Refuge.Restrooms.Search
       $('#lat').val() == undefined
       $('#long').val() == undefined
       if @_searchBar.val() == ""
+        @_searchBar.val(@searchCurrentLocationText)
         @_searchCurrentLocation()
       else
         @_searchQueryString()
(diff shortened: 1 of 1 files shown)


mi-wood:
I'd personally be in favor of a label vs using placeholder fields. There's much more liberty to be descriptive and it's permanent. Something like "Search by location (defaults to current location)". I don't think we're constricted with screen real estate or anything

mi-wood:
Also, sorry for the slow reply!

tkwidmer:
Yeah i'd prefer to use a label or a placeholder, but either of those would be wonderful! Thanks for this!

mi-wood:
@CullenIO Do you have plans to fix this up or should we close this for now?

mi-wood:
Going to close this for now, since it's been awhile

DeeDeeG:
(I notice this is closed, but this still might be something we could accept, since there are no merge conflicts last I checked.)

Alright, here's what this looks like for now (screen recording):

https://gfycat.com/FocusedMiserlyIlladopsis

Is there anything that should still be changed? 
- Should the search button have some text in it to explain what it will do?
- Label placed elsewhere?

tkwidmer:
I think the consensus was that we would rather this be a placeholder or a label vs changing the value of the field and putting this in as a value? @DeeDeeG

DeeDeeG:
I think I get what you're saying. It shouldn't end up in the search box as something users would accidentally submit. Correct me if I'm wrong.

FWIW, I'm clicking around, and I'm finding that it's just about impossible to accidentally search for the text "Searching by current location..." (I can't select it, I even tried to copy and paste it but when I try to select it, it kind of disappears and is replaced with light grey "Enter a query". It doesn't go through even when manually typed??)

The one placeholder in that box I've managed to get as a search, by accident, was "1 Embarcadero Center, San Francisco, CA ...etc..." but I can't figure out how to do it reliably.

PR #270, opened 2016-09-19
Improve contributing docs.

bellisk:
I had trouble getting started because of this, so adding it for future contributors.


mi-wood:
Thanks for the update and sorry you had trouble getting this running!

This looks to be a change in rails 4.2 where `rails s` defaults to `127.0.0.1` (https://stackoverflow.com/questions/27799260/rails-4-2-server-port-forwarding-on-vagrant-does-not-work).

Since this is a single use VM, it's entirely reasonable to bind to `0.0.0.0` instead of `10.0.2.15`. This will be resilient to and sort of vagrant networking change or defaults.

carolinux:
This worked for me:
```
vagrant ssh
cd /vagrant/
rails s --binding=0.0.0.0
```

PR #274, opened 2016-11-11
ruby: upgrade to 2.3.1

dunn:
The version can't be hard-coded in the Gemfile if we want to test
against multiple versions.


dunn:
Tests pass locally on both versions.

If you want I can add a commit bumping `.ruby-version` to 2.3.1 straight away, but I figured I'd be cautious.

cllns:
Does Heroku respect `.ruby-version` instead of the Gemfile? The site is hosted on Heroku, and I know they're the ones who ask for the Ruby version in the Gemfile. (I dislike it too)

Testing on multiple ruby versions might be overkill at this point, since there's only one instance of the site running (that I know of). I see it'd be useful to transition over, but I think that can be done by *switching* it to `2.3.1` and running the tests before deploying, instead of trying to test against both.

dunn:
> The site is hosted on Heroku, and I know they're the ones who ask for the Ruby version in the Gemfile.

Oh, forgot about that.  And ugh, looks like they only look at the Gemfile: https://devcenter.heroku.com/articles/ruby-support#ruby-versions

In that case I can close this or update it to instead change everything over to 2.3.1.

cllns:
Latter seems OK to me :) (FYI, not sure how much bandwidth @tkwidmer has right now, who does deploys, so it might take some time)

mi-wood:
Thanks @dunn for the changes! I'm going to take @cllns's word on this one and give it a 👍 I'll do a little local testing and try to get a deploy out soon!

PR #281, opened 2017-02-27
Landing page styling sc

vernistage:
For your consideration: I tweaked some of the main page design to blend elements with the graphic background on the main page & complement sizes of other elements. 
Thank you for creating this project.

### DIFF
--- a/app/assets/stylesheets/common.css.scss
+++ b/app/assets/stylesheets/common.css.scss
@@ -34,7 +34,7 @@
 
 .btn-purple {
   color: white;
-  background-color: rgba(65,50,107, .9);
+  background-color: transparentize($purple, 0.1);
   &:hover, &:focus {
     color: white
   }
--- a/app/assets/stylesheets/common.css.scss
+++ b/app/assets/stylesheets/common.css.scss
@@ -34,7 +34,7 @@
 
 .btn-purple {
   color: white;
-  background-color: $purple;
+  background-color: rgba(65,50,107, .9);
   &:hover, &:focus {
     color: white
   }
--- a/app/assets/stylesheets/search.css.scss
+++ b/app/assets/stylesheets/search.css.scss
@@ -15,7 +15,7 @@
 .header {
 
     .search-container {
-        margin: 10px 0px
+        margin-bottom: 5px;
     }
 
     .navbar-toggle:hover {
@@ -75,4 +75,3 @@
 //     color: black;
 //     border-color: $light-purple;
 // }
-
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/common.css.scss
```diff
@@ -34,7 +34,7 @@
 
 .btn-purple {
   color: white;
-  background-color: $purple;
+  background-color: rgba(65,50,107, .9);
```
Hkly: If that rbg is the same as `$purple` you can use [sass's transparentize](http://sass-lang.com/documentation/Sass/Script/Functions.html#transparentize-instance_method) so we can keep using the color variable for easy reference/editing.

mi-wood:
@vernistage Could we get some screenshots of before/after changes? Thanks so much for the PR!

tkwidmer:
@vernistage I'd love to see a before/ after screen shot. Just so we don't have to pull down your change and boot it up.

app/assets/stylesheets/common.css.scss
vernistage: Thanks so much! It is the same color. I don't know much about sass yet, but it looks way less finicky.

vernistage:
Sorry for the delay! I also wanted to transparentize the search form, but couldn't figure out how to do it without the background of the form turning purple. Here is after:
screen shot 2017-03-02 at 11 04 00 am

mi-wood:
@Hkly mind if I assign you to this one? my designer eye is not the best

Hkly:
Sure thing. Out of town right now, and can give a better look on Monday, but my initial thoughts are I like the same width for the title and search bar, but I think the add a bathroom button should be restored to its previous size too.

Hkly:
And if you can while you're in there, vertically center the button's text. It's just a tiny bit off center right now. I'm thinking line-height adjustment might fix, maybe?

Hkly:
I thought about the button width some more and full width would be great for mobile view, but typical "button size" for desktop view. Everything else looks pretty good overall though! Try out that sass function :)

tkwidmer:
+1 to returning the button size, but i'm good with it being full width on mobile, but on desktop it looks a little strange.

vernistage:
Awesome, sorry for the delay! I'm sure you all heard about SB6 in Texas. It has not been fun times. You all also inspired me to start using HAML instead of ERB, and it is amazing. I think I'll be able to contribute more.

mi-wood:
Thanks @vernistage and no worries about the delay! I'm going to go ahead and merge this now, since it looks like you've addressed folks comments :)

PR #283, opened 2017-02-27
change default language to English

vernistage:
Chrome thinks this website is in Danish for some reason, so I tried to change the default language to English in the Application Layout.


mi-wood:
Thanks @vernistage! I'd also like to have a future task of making the site more international accessible

PR #291, opened 2017-03-03
Update CONTRIBUTING.md

maecapozzi:
This PR clarifies how to get vagrant up and running locally.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ https://www.vagrantup.com/downloads.html
   * `cd  /vagrant/` to navigate to the refuge repo.
   * There is a rake wrapper to execute commands in refuge repo on the vagrant machine
   without using ssh. Run `rake vagrant:shell[command]`
-  * To start the rails server and bind it to the vagrant machine's IP address: `rake vagrant:shell['rails s -b 0.0.0.0']`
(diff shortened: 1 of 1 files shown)


cllns:
thank you!!

CONTRIBUTING.md
```diff
@@ -14,7 +14,7 @@ https://www.vagrantup.com/downloads.html
   * `cd  /vagrant/` to navigate to the refuge repo.
   * There is a rake wrapper to execute commands in refuge repo on the vagrant machine
   without using ssh. Run `rake vagrant:shell[command]`
-  * To start the rails server and bind it to the vagrant machine's IP address: `rake vagrant:shell['rails s -b 0.0.0.0']`
+  * To start the rails server bind it to the vagrant machine's IP address after running `vagrant ssh`, use: `rails s -b 0.0.0.0`. Otherwise, run: `rake vagrant:shell['rails s -b 0.0.0.0']`
```
mi-wood: Maybe this should be combined with the bullet point above. The thought of the wrapper is to help folks who don't need remote access (ex. if you're just trying to take a quick look at how a color changed). Could look something like this:
```
There are two ways to run commands in vagrant:

1. A local `rake` wrapper that allows direct execution on the machine... (go into depth about how)
2. Using `vagrant ssh` to gain access directly to the machine
```

PR #300, opened 2017-03-15
Fix for vagrant rake task bundler conflicts

dannypaz:
#### Description:
`rake vagrant:shell` was failing for some users because of conflicting bundler versions. This PR addresses that to allow the rake task to run under vagrants gemset/gems instead of a local users.

#### Changes:
- Added command to unset `RUBYLIB` per info [here](https://github.com/mitchellh/vagrant/issues/6158#issuecomment-135796049)
- Sets the ruby version on vagrant setup
- Installs a specific version of bundler (as [required](https://github.com/mitchellh/vagrant/issues/7193#issuecomment-204309088) by vagrant)

### DIFF
--- a/.rake/vagrant.rb
+++ b/.rake/vagrant.rb
@@ -1,9 +1,14 @@
-vagrant_path = 'PATH=/home/vagrant/.rbenv/shims:$PATH'
-
 namespace :vagrant do
   task :shell, [:command] do |_t, args|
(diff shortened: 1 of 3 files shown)


dannypaz:
@tkwidmer let me know if more documentation is needed for these changes

dannypaz:
woopwoop! thanks for reviewing @mi-wood

PR #304, opened 2017-03-15
update copyright

mi-wood:
@tkwidmer just updating the year :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Channel: #refugerestrooms
 
 ## License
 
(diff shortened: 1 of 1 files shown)


cllns:
FYI, you only need the first year, it doesn't need an end year. http://www.photoattorney.com/qa-what-year-should-i-use-for-my-copyright-notice/ 

Also probably don't need the copyright statement in the README at all? Not many big projects do it

mi-wood:
:woman_shrugging: you're probably right. I just saw an inconsistency :P

Issue #287, opened 2017-03-02
Upgrade Rails to 5.1


tkwidmer:
Rails has been upgrade. We need to try to stay current.

### Upgrade from 4.2 -> 5.0
http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-2-to-rails-5-0

### Upgrade from Rails 5.0 -> 5.1
http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-0-to-rails-5-1

dannypaz:
I've been doing a little bit of research on our upgrade path. Here is some info

#### Gems to update

- rails 
- coffee-rails 
- dotenv-rails 
- quiet_assets (Removal)
  - See more info [here... deprecation](https://github.com/evrone/quiet_assets#deprecation)
- devise 
- jquery-rails
- turbolinks
- rack-contrib (See below)

#### Blockers

`rack-contrib` does not work with Rack 2, which is required for Rails 5. See more [here](https://github.com/rack/rack-contrib/pull/129). We can either fork this library and apply the patch OR we remove Grape for Rails 5 API.

dannypaz:
@tkwidmer I have absolutely no idea what files can be removed from the repo, so i'll probably need some help in identifying these

mi-wood:
Deployed at https://refuge-production.herokuapp.com/ !!

tkwidmer:
Now just to get from 5.0 to 5.1!

mi-wood:
fyi, this is currently blocked due to a transitive dependency between rails and devise. Will probably just wait for a release of devise

PR #307, opened 2017-03-17
Rails 5 Upgrade

dannypaz:
Please view the Rails 5 issue https://github.com/RefugeRestrooms/refugerestrooms/issues/287 for details or the supporting [documentation](http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-2-to-rails-5-0) for more info.

#### Description

This PR upgrades the refugerestrooms application to Rails 5 (awww yeee).

#### Changes
- Upgraded a boatload of gems to satisfy `railties` and `rake` requirements
- Removed `rack-contrib` and monkey patched `Rack::JSONP` support (temp fix)
- Removed deprecation for  `before_filter` and changed to  `before_action`
- Removed deprecation in passing `Rack::Cors` as a string to middleware
- Removed `top_cities` code from PagesControllers (part 2 of https://github.com/RefugeRestrooms/refugerestrooms/pull/327)

#### TODO
- ~~Fix PagesController (not able to load site)~~
- ~~remove requires from asset pipeline in favor of using @imports w/ rails-sass (please see rails-sass documentation for more info)~~
- ~~bump application version numbers (or major versions)~~
- ~~Make sure test suite works~~
- ~~Update rails config files w/ `bin/rails rails:update`~~
- ~~Update/Upgrade/Import remaining gems from Rails 5 gemfile~~
- Try to deploy to heroku w/ Cedar-14 (@teagan)

Test Suite:
```
# Rspec suite
Finished in 2.27 seconds (files took 6.29 seconds to load)
48 examples, 0 failures

Randomized with seed 60083

Coverage report generated for RSpec to /Users/dev/Documents/refugerestrooms/coverage. 453 / 520 LOC (87.12%) covered.
```

```
# Cucumber tests
9 scenarios (9 passed)
26 steps (26 passed)
0m34.397s
```

#### Screenshot

screen shot 2017-05-09 at 11 24 49 pm


cllns:
could we use https://github.com/crohr/rack-jsonp instead of monkey patching JSONP ourselves?

dannypaz:
@cllns We can. FWIW, It is the same exact code -_-

mi-wood:
@dannypaz thanks for submitting this! I'll try to get to it soon. @cllns since I'm pretty out of sync with rails these days, do you mind doing a review when you get a chance? @tkwidmer thinking we should wait to fix the maps API issues we're having before this? Just because that seems important to fix before putting more changes in. Thoughts?

dannypazz:
@mi-wood I've updated the PR w/ current develop so we can start testing. @cllns let me know if you have any questions RE: the changes made for Rails 5 (it was pretty straight forward).

@mi-wood is there a PR or issue up for the maps API issues?

dannypazz:
Also, FWIW... I still cannot figure out this testing error as I receive the error on develop and this branch. Can someone take a look at that?

dannypazz:
@mi-wood Im taking a look at the maps issues w/ this PR and it looks like the only thing that is broken is the text search on the home page.

You can view the project on my test container here: https://secure-coast-64861.herokuapp.com/

mi-wood:
@dannypaz I added an API key for the maps, so that should be fine. The testing error is actually the same one as the homepage search. I wouldn't block this PR on it, although we should fix it before updating the app to point to the new deploy.

mi-wood:
I'm mostly ok with this PR. Lots of config/deprecated syntax changes it seems. I'd just like a second pair of eyes before merging it.

config/cable.yml
```diff
@@ -0,0 +1,9 @@
+development:
```
cllns: What do you think about removing this file? If we want to add ActionCable, we can add this file then

config/initializers/application_controller_renderer.rb
```diff
@@ -0,0 +1,6 @@
+# Be sure to restart your server when you modify this file.
```
cllns: What do you think about removing this file? I think it's only for ActionCable? If we want to add ActionCable, we can add this file then :)

config/initializers/jsonp.rb
```diff
@@ -0,0 +1,80 @@
+require 'rack'
```
cllns: I'd prefer to use https://github.com/crohr/rack-jsonp instead :) 

If we want to have the code here for some reason, we should copy [their specs](https://github.com/crohr/rack-jsonp/blob/master/spec/rack_jsonp_spec.rb) into our codebase too

tkwidmer:
I'm going to defer to @cllns and @mi-wood when it comes to this PR.

DeeDeeG:
This PR makes some small css style changes, I infer these are unintentional. Probably just comes down to which stylesheets end up imported nearer to the top of the compiled application.scss, (less prioritized) vs toward the bottom of the css file (more prioritized), or something like that.

I'm looking into what changed in particular, and how to prevent it from changing.

Example: the search/nav buttons at the top of most pages (other than the home page).

DeeDeeG:
Copying and pasting application.scss from develop into this branch solves all styling issues (that I can see on first glance) from my previous comment. I think it can/should be reverted.

A lot of the original file is wrapped in /* */ which looks like comment-out syntax. But all those require statements do get executed. There isn't any benefit I am aware of to moving to @import statements, they work the same, right?

app/assets/stylesheets/application.scss
```diff
@@ -8,19 +8,19 @@
...
- *= require top_cities
- *= require common
- *= require restrooms/new
- *= require splash
+ * @import 'myfile'
  */
 
 @import 'refuge_core';
+@import 'mobile';
+@import 'active_admin';
```
DeeDeeG: Importing active_admin causes styles to be overridden that probably shouldn't be.

I don't think active_admin styles have been on for a while. (Maybe it should be removed from the project?)

tkwidmer: active admin is the admin panel for deleting bad data, although its not very used. the people who were doing some of that have largely not been doing anything for a while.

DeeDeeG: I could do some of that, and/or other administrative/data-entry stuff, if that works for you/core maintainers.

DeeDeeG:
It would be nice to see this added, as far as I'm concerned. I did test it (with vagrant) and it seems to work well.

I believe the TODO is:

- add Rspec tests (aka "specs") for Rack-jsonp to our repository, (or instead, just add rack-jsonp to our project as a gem via our Gemfile)
- remove files related to ActionCable,
- and consider undoing functional changes to the SCSS files. (However, renaming _mobile.*css.scss* -> _mobile.*scss* is fine and gets rid of some warnings, so I'm all for that.)

dannypaz:
@DeeDeeG +1 to all the suggestions. FWIW, I would much rather switch to `rack-cors` than install `rack-jsonp`.

config/initializers/jsonp.rb
dannypaz: +1. I would rather not install the gem, specifically because 1. it is dead and 2. the code isn't very complex. @DeeDeeG made a suggestion below to use `rack-cors` which hopefully will be a better alternative

dannypaz:
@DeeDeeG @cllns @tkwidmer Let me know if this sounds good.

For now, I will install `rack-jsonp` in this branch. Moving forward, i've created this issue to upgrade to `rack-cors` since this change is not required to move forward with upgrading to rails (although we should make the change eventually).

Issue can be viewed here: https://github.com/RefugeRestrooms/refugerestrooms/issues/341

config/initializers/jsonp.rb
dannypaz: Check out my comment below @cllns

DeeDeeG:
Looks good to me. 

I poked around on all the user-facing pages, and (as far as I can tell) they all function and appear identical to develop branch.

DeeDeeG:
As a minor thing, we are using the "restrooms/new" styles twice. Two very easy ways to fix this: 
- we can drop "@import restrooms/new" from the beginning of `restrooms.scss` 
OR
- rename `new.scss` to `_new.scss`
(the underscore turns it into a "partial", which Sass knows to not to send to users, on its own. Partials only get activated if they are called for with @import. partials actually get rendered into the file(s) that import(s) them, before the server sends the processed CSS out to users. Since `new.scss` is currently imported by `restrooms.scss`, we would get it, and only get it once, if new is converted to a partial.)

(on a tangent, this is for another issue or PR: I think we might be double-importing SCSS files elsewhere in the repo, now that I take a closer look. It would be nice to tidy that up, and might help with performance especially during pageload/page refreshes.)

tkwidmer:
I dont know if we should commit this until we figured out why Travis isn't able to run it

DeeDeeG:
I don't strictly speaking know why this happens, as in why bundler has this as an issue that can happen, because it seems like a weird design decision, but I do know what the problem is/how to fix it. It happened to me once. Just need to get an updated Gemfile.lock, which can even be done by deleting the old Gemfile.lock and then running `bundle install`. (It's supposed to be a security mechanism, but I can't image a scenario where it would be helpful. It's probably just that we use version control, and accidentally mismatching the Gemfile.lock to the current Gemfile happens much more easily with version control than without. :shrug:) So it genuinely confuses me why this is a thing, but I can describe the fix.

For our situation here (background info first):

Travis says the Gemfile.lock is out of date and needs to be checked in with "version control", aka git for our pusposes.

Gemfile.lock is a file Ruby (or bundler?) makes, keeping track of the list of dependencies (gems) that were pulled in (during the last successful run, i.e. finished without errors) of `bundle install`. So to fix it, running vagrant up will call the setup script, which runs bundler... Which produces the updated Gemfile.lock.

#### That said, the steps to fix this are:

1. Run `vagrant up` (This will slightly change Gemfile.lock, because we added the `rack-jsonp` gem since last time a copy of Gemfile.lock was uploaded to our repo)
2. Find Gemfile.lock in the working "refugerestrooms" git folder, and add the changed version to be tracked by git (`git add Gemfile.lock`)
3. Commit the change and push to github. (`git commit [other commandline parameters here]; git push`)

I know I might be overexplaining, but it's a safer bet than underexplaining, so I tend to lean toward overexplaining.

DeeDeeG:
Also, here's my updated Gemfile.lock, which could just be copy-pasted and committed, if that's easiest.

Link to Gemfile.lock: https://gist.github.com/DeeDeeG/c199a73a6a9b332d99afaf9f30f3198f

dannypaz:
@DeeDeeG i'm so embarrassed. I had the `gemfile.lock` changes still stashed. They are now commited :-).

DeeDeeG:
@dannypaz it happens to the best of us. :)

So as expected, Travis is working again. :tada: No build failures related to this PR.

(Digging into what is unrelatedly failing: Interestingly, after a quick glance, we don't seem to be failing on the Google Maps unauthorized key issue right now. We only seem to be failing on a test for top cities, which makes sense knowing that the feature was completely removed. If we remove the top cities tests, we might be back to having passing builds!)

dannypaz:
For record keeping, here is the PR @DeeDeeG is referencing: https://github.com/RefugeRestrooms/refugerestrooms/pull/327

I've removed the offending test, so we should have a passing build now. Was this the correct solution? @DeeDeeG

DeeDeeG:
That's definitely the right part of the file to delete, but still a couple of the top cities tests are left in the commit. (This can be fixed again without adding extra commits, by making changes, `git add`ing the changed file, and using `git commit --amend`, if you like.)

There's a lot of nested "do ... end" in the file, so it's not very readable to start with, and therefore a bit hard to edit?

Anyways: The only way I know to do this right, by hand, is to delete the innermost "do ... end" block, and the next innermost "do ... end" block, then move out one level, delete some more blocks one "do ... end" at a time, until the whole top cities test is gone.

 I treat "do" as if it was "{", and "end" as if it was "}" because they mark blocks of code like I would expect brackets to.

dannypaz:
@DeeDeeG Ahh. OK I think I understand correctly now. I think these tests and related controller code should have been removed with https://github.com/RefugeRestrooms/refugerestrooms/pull/327 but that is OK.

I've added a new screenshot to the description and have added `removed top_cities code` to one of the changes.

dannypaz:
TY for the help. Looks like we are all green.

mi-wood:
Awesome work ya'll! I should have some time to take a look at this today

DeeDeeG:
I expect to be busy soon, and I'm new to slack, but sure, I can see what it's all about. Thank you.

mi-wood:
I'm mikena@woodmail.me if you want to shoot me your email, I'll add you :)

tkwidmer:
🎉  thank you so much @dannypaz

PR #310, opened 2017-03-20
fix setup with .ruby_version newlines

### LINKED ISSUE #309: rbenv issue on vagrant up
On first-time start of the vagrant box, rbenv fails to install ruby 2.3.1:

```
==> default: installing rbenv...
==> default: Cloning into '.rbenv'...
==> default: Cloning into '/home/vagrant/.rbenv/plugins/ruby-build'...
==> default: install ruby 2.3.1
==> default: ruby-build: definition not found: 2.3.1
==> default: See all available versions with `rbenv install --list'.
==> default: If the version you need is missing, try upgrading ruby-build
==> default: :
==> default:   cd /home/vagrant/.rbenv/plugins/ruby-build && git pull && cd -
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

kurtcarpenter:
Previously, setup_vagrant.sh would fail when newlines were present on .ruby-version. Now the newlines are stripped and ruby installation works correctly.

Fixes #309

### DIFF
--- a/setup/setup_vagrant.sh
+++ b/setup/setup_vagrant.sh
@@ -54,7 +54,7 @@ eval "$(rbenv init -)"
 export PATH="$HOME/.rbenv/plugins/ruby-build/bin:$PATH"
 
 # Install ruby
-ruby_version=`cat $REFUGE_PATH/.ruby-version`
+ruby_version=$(cat $REFUGE_PATH/.ruby-version | tr -d '\n\r')
 if rbenv versions | grep $ruby_version; then
   echo 'ruby '$ruby_version' installed, skipping...'
 else


setup/setup_vagrant.sh
```diff
@@ -54,7 +54,7 @@ eval "$(rbenv init -)"
 export PATH="$HOME/.rbenv/plugins/ruby-build/bin:$PATH"
 
 # Install ruby
-ruby_version=`cat $REFUGE_PATH/.ruby-version`
+ruby_version=$(cat $REFUGE_PATH/.ruby-version | tr -d '\n\r')
```
mi-wood: Kinda neat trick I just learned: `xargs < $REFUGE_PATH/.ruby-version` will do this too

PR #311, opened 2017-03-24
Update seed file sc

### LINKED ISSUE #303: Update export.csv data
`db/export.csv` hasn't been updated in almost 3 years, so the data _may_ be stale. We should make sure these locations are correct/up to date.

vernistage:
# Context
- Fixes #303 
- Update data in csv seed file

# Summary of Changes
- 12, Mission Creek Cafe is permanently closed per Yelp & Google.
- 14, Rainbow Grocery Co-op is listed as "Rainbow Grocery"
- 17, Cafe Roma is listed as "Caffe Roma"
- 18, Maxfield's House of Caffine is "Maxfield's House of Caffeine"
- 19, Cafe Espresso is permanently closed per Yelp & Google
- 21, Muddy Water's is "Muddy's Coffee House" and is now at 1304 Valencia St San Francisco, CA 94110
- 40, Philz Coffee there is no longer a Philz Coffee at 3901 18th St
- 38, Bang San Thai Restuarant is just "Bang San Thai"
- 66, New College is permanently closed
- 42, Sparky's Diner is permanently closed
- 50, Muddy Water's Coffee is Muddy Waters Coffee House
- 51, GLBT Historical Society is now at 989 Market St, San Francisco, CA 94103
- 53, LGBT Community Center is "The SF LGBT Center"
- 68, maggie mudd ice cream is permanently closed

# Checklist

- [N/A] Tested Mobile Responsiveness
- [N/A] Added Unit Tests
- [ ] CI Passes

# Screenshots

## Before
screen shot 2017-03-24 at 3 30 03 pm

## After
screen shot 2017-03-24 at 3 29 42 pm


mi-wood:
@vernistage I'm all for this change. Can you just peak into why the build is failing on it? Thanks!

vernistage:
Sure, The only thing I can think of is that one of the lines may contain a rogue comma? The csv looked fine as a complete table in my forked repo's github preview. If anyone else has any ideas, lmk.

vernistage:
Here's a preview of the data as a table: https://github.com/vernistage/refugerestrooms/blob/update-seed-file-SC/db/export.csv. Do you think the build failed because it was a seed file?

DeeDeeG:
I looked at the Travis logs, and I think there is a test expecting the Mission Creek Cafe to turn up. Since this PR's updates delete the cafe, and because the cafe does not exist anymore, that test is no longer functional. I think the test should probably be recoded to use some restroom location that is unlikely to get removed in real life, if possible.

(More specifically: Travis tries to click the "Request an edit to this page" link, or click something to make a "generic contact", from within the Mission Creek Cafe restroom page. But I'm not sure that page even exists if the restroom data is gone. So those two tests fail.)

dannypazz:
@DeeDeeG It sounds like we should be stubbing this out in the test so it isn't dependent on the seed file. Is that possible with our current setup?

DeeDeeG:
Hmm, I'm not so familiar with this, I'm kind of new to the subject, actually.

(my inclination is to say that anything is possible. If I looked into / knew how the tests were written, I'd know whether we could set aside a "dummy" restroom for tests only. Then, regardless of how that happens, it'd be nice to remove that from the production database, or make sure it doesn't get there in the first place. As a last resort, I guess we could put a restroom in the middle of the ocean in the seed, so users won't likely notice it; or taking another easy way out, program the test to use a real restroom unlikely to be removed IRL.)

I THINK that when we test locally, and when Travis runs tests, right now, the ONLY restroom are those in the seed file... But I'm not 100% on that,  more like 10%. 

(If anyone else has these answers, obviously the real answers would be good to know. I'll double-check myself when I have time.)

DeeDeeG:
This is the commit where these two tests are added: https://github.com/RefugeRestrooms/refugerestrooms/commit/dc6c7c796067932d4f0c405e5fe28f77f9cd55a3

And these are all the commits that touched that test: https://github.com/RefugeRestrooms/refugerestrooms/commits/develop/features/contact.feature

This should be a good starting point for understanding how the test works, how to rewrite, etc. also since @mi-wood made the test, they might have a better idea how to fix it. (I'm still looking into it, I just don't know what I'm doing yet.)

mi-wood:
Seems like the way to do this would be to use a fixture or something rather than relying on the seed. I can't find something that correlates to cucumber exactly, but my quick search finds that it should be a combination of a `Before` hook and a helper method to add the restroom. Possibly even a cleanup step to get rid of it.

I can try to take a look at adding this during the week, unless someone else wants to pick it up!

DeeDeeG:
This looks like a dummy restroom, and thus a way to fix our problem here. Found this while writing the docs for RSpec. 

https://github.com/RefugeRestrooms/refugerestrooms/blob/develop/spec/factories/restrooms.rb

I admit I don't 100% know what I'm looking at. But a naive assumption is that we can simply test against the Moonlight Cafe and it should work.

DeeDeeG:
We also have this, which looks like a good example of how to set this up directly in RSpec. Copy and pasting this, and trial and error, and a bit of reading the docs, should give us a working RSpec test (even without knowing exactly what we're doing), optimistically speaking.

https://github.com/RefugeRestrooms/refugerestrooms/blob/e379eb8057c68173111b6827df0dbbec8be658f6/spec/spec_helper.rb#L50

```
Geocoder.configure(:lookup => :test)

Geocoder::Lookup::Test.add_stub(
  "123 Example St., San Francisco, CA, US", [
    {
      "latitude"     => 37.7749295,
      "longitude"    => -122.41941550000001,
      "address"      => "123 Example St., San Francisco, CA, USA",
      "state"        => "California",
      "state_code"   => "CA",
      "country"      => "United States",
      "country_code" => "US"
    }
  ]
)
```

DeeDeeG:
Hi all,

As far as I see it, there are three ways forward on this PR. In no particular order, we can:

- Change the current tests by tweaking the hard-coded restroom it tests for.

- Update the tests by having the tests create a stub restroom entry before the tests, then destroy said stub after running.

- Not move forward and close the PR. ( :frowning_face: )

I note that this file is only used during testing. Stubbing out a separate test-only restroom seems like lots of work to me, considering these are already made for testing and serve mostly the same purpose as a stub would.

I would personally go with tweaking the hard-coded restroom the tests use, for simplicity's sake. Updating the test, if still desired, could be moved into a separate issue/PR.

mi-wood:
I'm going to close this for now. If someone wants to pick it up again, I'm happy to revisit. There have been quite a few changes since the start of this, so it might need some tlc for an update and rebase.

Issue #302, opened 2017-03-15
`last_comment` is deprecated.  Please use `last_description` instead.


dannypaz:
This occurs due to the rspec version. We will need to update rspec to remove these deprecation warnings when running rails server in VM.

More info [here](https://github.com/ruby/rake/issues/133#issuecomment-217581285)

PR #315, opened 2017-03-27
Mi wood/deprecations

mi-wood:
closes https://github.com/RefugeRestrooms/refugerestrooms/issues/302
upgrades 'rails-rspec' to get rid of deprecation. dependency upgrade of 'rails-sass' caused some extension deprecations that I fixed with some file renames


tkwidmer:
lgtm

Issue #297, opened 2017-03-14
Google Maps API needs key?


lgdean:
Running the app locally, I searched in San Francisco, clicked on one of the results in the list (http://localhost:3000/restrooms/10), and got a page with info about the bathroom, and an error from Google instead of a map.

In the JS console, I found a `Google Maps API error: MissingKeyMapError` error, and two warnings.

See attached screenshot.
<img width="1433" alt="screen shot 2017-03-13 at 8 39 03 pm" src="https://cloud.githubusercontent.com/assets/371459/23884971/7ae15ed6-082d-11e7-9fd0-81233dc049b1.png">

Chrome 56 on a Mac, in case it matters.

lgdean:
My friend tried too, and was sometimes getting the error, and sometimes not.

The wifi is too slow, at the moment, to see whether the same problem happens in production.

dannypaz:
@lgdean This looks to be an issue w/ the data from `db/export.csv`. Philz Coffee and the associated data [no longer exist at 3901 18th st](https://www.yelp.com/biz/caf%C3%A9-ub-san-francisco-2). 

The 2 warnings from Google Maps API should not affect the features of the application.

dannypaz:
@lgdean I've made a separate issue to address this https://github.com/RefugeRestrooms/refugerestrooms/issues/303

mi-wood:
I'm heading on vacation until next week, but if this is still up when I get back, I'll take a stab at this along with #299

mi-wood:
Taking a first glance and it seems like the last time we deployed was 12 months ago and as of last June they started requiring an API key for the JS API: https://maps-apis.googleblog.com/2016/06/building-for-scale-updates-to-google.html

We'll need to follow this and distribute the key: https://developers.google.com/maps/documentation/javascript/get-api-key

Haven't gone through the setup yet, but it seems like it has to be updated here: https://github.com/RefugeRestrooms/refugerestrooms/blob/e379eb8057c68173111b6827df0dbbec8be658f6/app/views/layouts/_head.html.haml#L9

It doesn't matter how well encrypted/stored it is, since you could get it just by querying our app, but it should be thrown into a config. That being said, I'm not seeing if there's an easy way to stop someone from rate limiting rate limiting us unless we do the queries server side or maybe google has IP based rate limiting built in for us?

Anyways, mostly seems administrative (making a google account and creating the key).

tkwidmer:
[...]

Issue #288, opened 2017-03-02
Upgrade Heroku Ceder Stack


tkwidmer:
# Context

Heroku has depricated Ceder-10 which is what RefugeRestrooms is on. We need to update heroku to Ceder-14.

https://devcenter.heroku.com/articles/cedar-14-migration

<img width="670" alt="screen shot 2017-03-02 at 11 30 17 am" src="https://cloud.githubusercontent.com/assets/5265597/23523410/a30b8dfa-ff3b-11e6-8edb-d5c1ff2a930e.png">

tkwidmer:
I'm planning to try to take this in the upcoming hack day

tkwidmer:
https://github.com/RefugeRestrooms/refugerestrooms/issues/297 Needs to be resolved before we can deploy again.

tkwidmer:
Something about Piwik or our setup of it is blocking the application from functioning on the new Heroku Cedar-14. I reverted some of the work in #206 and was able to get the app functioning for the most part. The maps are still broken on https://refuge-staging-14.herokuapp.com/, but it's in better condition. I'm using a test branch to try to do this work. I'll have to disable Piwik to get deploys back working.

tkwidmer:
Now that #327 is merged, and #322 is closed. I think we may be good to go on the switch over

mi-wood:
Yay this is deployed!

Issue #260, opened 2016-03-24
Release 1.8.0


tkwidmer:
Teagan (6):
      Merge pull request #238 from aitchabee/patch-1 @aitchabee 
      Merge pull request #243 from azelma/develop @azelma
      Merge pull request #245 from cllns/strong-params-on-restroom @cllns 
      Merge pull request #248 from azelma/changing-tables @azelma
      Merge pull request #250 from azelma/static-pages @azelma
      Merge pull request #254 from RefugeRestrooms/master

mi-wood (1):
      Merge pull request #240 from brntbeer/patch-1 @brntbeer

tkwidmer:
I have to upgrade the Ceder stack on both staging and production to be able to release these changes.

tkwidmer:
I'm going to close this, since we will probably skip to 1.9.0 when we are able to deploy again.

PR #316, opened 2017-03-27
Mi wood/fix heroku

mi-wood:
Some miscellaneous changes to get heroku working. @tkwidmer I was getting piwik timeouts, is this just an API/environment thing? Can easily add that back in, but it just froze the page :(

- Provisioned a google maps key. I don't see an easy way to do this other than have a public key. It's a 25,000 request limit per 24 hours. I'd like to see if we can set our own rate limit per IP or something.
- Had to change a deprecated rails config
- Some changing_table stuff was off and breaking queries, so I fixed that (also had to db:migrate)

### DIFF
--- a/app/admin/restroom.rb
+++ b/app/admin/restroom.rb
@@ -1,5 +1,5 @@
 ActiveAdmin.register Restroom do
 
-permit_params :name, :street, :city, :state, :accessible, :unisex, :directions,
+permit_params :name, :street, :city, :state, :accessible, :changing_table, :unisex, :directions,
               :comment, :latitude, :longitude, :country
 end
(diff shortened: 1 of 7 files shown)


mi-wood:
closes https://github.com/RefugeRestrooms/refugerestrooms/issues/297 
closes https://github.com/RefugeRestrooms/refugerestrooms/issues/288
closes https://github.com/RefugeRestrooms/refugerestrooms/issues/260
not sure if I could have done that differently...

tkwidmer:
This looks good to me.

Although it looks like one of the CI Jobs didn't pass. I'm not sure why there are two jobs now? I think this puts us in a better state, and I don't think this causes the build break. So I'm going to merge this. <3 <3 @mi-wood 

https://travis-ci.org/RefugeRestrooms/refugerestrooms/builds/215382110

PR #351, opened 2017-06-13
Disable geolocation spinner on location error, notify user

### LINKED ISSUE #264: Location Services Disabled; Tell them they need to enable.

kurtcarpenter:
# Context
- Fixes #264 
- Previously, if the geolocation was unavailable, the spinner would continue and clicking it would do nothing

# Summary of Changes

- Now, the spinner stops and the original location arrow is restored when geolocation fails; additionally, an alert pops up asking the user to refresh the page and enable location services (on first fail, and each click of the button)
- Additionally, two issues involving incorrect Promise usage are addressed in the geocoder. These are the misspelling of "reject", and use of "fail" instead of the correct "reject" (so it added a handler, instead of actually rejecting the Promise).

This PR uses the browser default alert window, for lack of another way to message the user. If anyone has prettier suggestions, I would be happy to address.


app/assets/javascripts/views/restrooms/search.coffee
```diff
@@ -62,9 +62,13 @@ class Refuge.Restrooms.Search
 
 
   _searchCurrentLocation: =>
-    @_geocoder.getCurrentLocation().then (currentCoords) =>
-      @_updateForm(currentCoords.lat, currentCoords.long, "Current Location")
-      @_form.submit()
+    @_geocoder.getCurrentLocation()
+    .then (currentCoords) =>
```
Hkly: This and the `.then` on line 69 should be indented one more level

kurtcarpenter: Is that a stylistic or syntactic choice? I believe this works properly having tested it and referring to examples such as https://stackoverflow.com/questions/27595933/coffeescript-promise-chaining-with-function-definition#answer-27595943

Thanks for reviewing!

Hkly: Ah if the rest of the code base doesn't indent the `.then` then you should leave it, but the next line below the `.then`s are now double indented :P My personal preference is to indent the `.then` so it's more obvious that it's chaining off the line above it.

mi-wood:
@kurtcarpenter I had to do some surgery on the main branches. Could you re-checkout develop and add these changes, please? @Hkly if you're good with these changes, I am!

Hkly:
LGTM 👍

kurtcarpenter:
FYI this has been updated with latest from `develop`

mi-wood:
Closed by https://github.com/RefugeRestrooms/refugerestrooms/pull/359

PR #353, opened 2017-06-28
Align icons in restroom list to the right

### LINKED ISSUE #320: Allign accessibilty filter icons
Icons get aligned with the start of the distance text. This means that they aren't aligned between listing, which seems a little odd to me. Maybe we can align right instead.

Hkly:
# Context
- Fixes #320
- Icons were aligned against distance text length and not uniform between listings.

# Summary of Changes

- Align icons and distance text to the right
- Display icons inline-block

# Screenshots

## Before
screen shot 2017-06-28 at 3 29 59 pm
## After
screen shot 2017-06-28 at 3 28 49 pm

### DIFF
--- a/app/assets/stylesheets/restrooms.scss
+++ b/app/assets/stylesheets/restrooms.scss
@@ -55,7 +55,7 @@
         text-align: right;
         font-size: 15px;
         font-family: Arial;
-        padding: 5px;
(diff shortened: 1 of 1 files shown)


mi-wood:
👍  Thanks @Hkly !

PR #369, opened 2017-08-19
Fix preview not loading after pressing guess button

### LINKED ISSUE #366: Preview does not load with "Guess current location"
The preview works great if I fill in an address, but "Guess current location" causes the button to not work. I'll take a quick peak as to why. Still going to deploy to master, since this is a minor issue

stardust66:
# Context
- Fixes #366
- Make preview load when guessing location as well. Also fix clicking guess button disabling preview button.

# Summary of Changes

- Move code that unhides the map to the updateMap function, so that it gets called when the guess button is clicked too.
- Add rebind function. See explanation in [commit message](https://github.com/RefugeRestrooms/refugerestrooms/commit/797f45e49a3ac88cd09b1b957c46729b200685d5). This fixes not being able to click the guess button twice, and that clicking the guess button disables the preview button.

# Checklist

- [ ] Added Unit Tests
- [ ] CI Passes

### DIFF
--- a/app/assets/javascripts/views/restrooms/new.coffee
+++ b/app/assets/javascripts/views/restrooms/new.coffee
@@ -30,6 +30,7 @@ class Refuge.Restrooms.NewRestroomForm
             @_getNewForm(coords).then (data, textStatus) =>
               console.log data
               $('.form-container').html(data).hide().fadeIn()
+              @_rebind()
               @_requestNearbyRestrooms(coords)
               @_updateMap(coords)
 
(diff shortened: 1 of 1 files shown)


DeeDeeG:
I found out that the Preview button fails when the "Restroom Name" field (on the new restroom form) isn't empty. (This happens because we get a bad response from the Maps API).

In the test file, [preview_steps.rb](https://github.com/RefugeRestrooms/refugerestrooms/blob/aa4c80e3bebd23dc6c16829971fa051ba5acd0d6/features/step_definitions/preview_steps.rb#L4), the restroom name is filled out (with the text "Vancouver restroom" for the name).

So, given that the name field isn't empty, and due to the Maps API not liking the request we send out, at the moment the Geolocation fails and the test subsequently fails.

I looked into it a bit further, and the Restroom Name value gets passed to Google as part of the request we send out. I assume if that is not sent, we'll get a good response.

DeeDeeG:
Seems like changing this line should fix it: [new.coffee, line 44 or so](https://github.com/stardust66/refugerestrooms/commit/0deb564361be8b286229696e733c785647e83112#diff-c3d073642428cc5fed25cd969aa8fbb2R44)

address = "~~#{name},~~ #{street}, #{city}, #{state}, #{country}"

mi-wood:
Merging to a separate branch so I can fix the other issue

PR #370, opened 2017-08-19
Change form input color to black

stardust66:


app/assets/stylesheets/restrooms/new.scss
```diff
@@ -1,3 +1,7 @@
+.form-control{
```
mi-wood: I think we can do this globally and not just in "new". The location lookup also displays gray test.

stardust66: Alright I moved it.

Issue #251, opened 2015-11-08
Adding highway rest areas to database


hissingpanda:
An android user was wondering how to add some rest stops in MI as most rest areas don't have street addresses. An option to add Lat/Lng instead of a street address when submitting a restroom would be a way of solving this.

tkwidmer:
Right now these are hidden fields, we could make them unhidden. And maybe design the experiance a little bit better. Where there could be a link that says, I want to enter the Lat/Long. Alternatively, on the mobile web, I think you can guess your current location which fills in the lat / long -- address fields as well.

hissingpanda:
The current location is a great option, but that particular user said they didn't always have time to log the restrooms at that very instant. Reception probably isn't the greatest in more rural rest stops too. A select box to choose between address and Lat/Long is an option, where if you select Lat/Long it unhides those fields. That way most people could just ignore it. I've noticed a lot of college campus buildings don't have street addresses either.

hissingpanda:
A drag and drop option after submitting is also mentioned in #249, and seems like a good alternative to me.

tkwidmer:
@hissingpanda Drag and Drop functionality would be great.

Issue #249, opened 2015-10-18
Users to own a bathroom with the ability to edit them? 


t-walker:
While submitting restrooms, one of the addresses didn't place properly and I couldn't edit them. Would it be worth implementing a login system so users can own a bathroom that they submit and edit at a later time?

drdpedroso:
What if we leave it all open? For any user to edit?

t-walker:
That works as well. Perhaps suggested edits could go into a queue for admin approval?

drdpedroso:
Nice! I'll try to implement that in my free time

tkwidmer:
@t-walker @drdpedroso There is already a ticket / PR for this #198

tkwidmer:
But its been stagnet for a while

t-walker:
I added a bunch of bathrooms and since my university refuses to publish the actual street address of the buildings a lot of them are just pinned to Pullman and I'm not sure how to change that. Perhaps before committing to the db you allow people to see where it is pinned and then drag and drop?

tkwidmer:
@t-walker I think that also sounds like a good solution.

mi-wood:
closing since this is a dupe of https://github.com/RefugeRestrooms/refugerestrooms/issues/126

PR #374, opened 2017-09-04
Drag and drop map marker when submitting new restroom

stardust66:
# Context
- Mentioned in #249 and #251.
- Users can now drag the map marker to the desired location in the map preview, and the form will be updated with the new address obtained by reverse geocoding the coordinates of the marker.

Note: the map will disappear after dragging the marker once without my changes in #369.

# Summary of Changes

- Modify `initMap()` to take the arguments `draggable` and `callback`, which make the map marker draggable. `draggable` defaults to `false`. This can also be useful for potential editing functionalities.
- Add `window.Maps.reloadDraggable(map, callback)`, which can be used to update a map with draggable markers.
- Put the code for updating the form in the `_updateForm` method, as the same code gets called multiple times.

# Checklist

- [ ] Added Unit Tests
- [ ] CI Passes

# Screenshots
drag

### DIFF
--- a/app/assets/javascripts/maps.js
+++ b/app/assets/javascripts/maps.js
@@ -4,7 +4,7 @@ function initMap(x, y, image, draggable, callback){
 	image = typeof image !== 'undefined' ? image : currentLocationImage;
 
 	// Draggable defaults to false
-	draggable = (draggable === undefined) ? false : true;
+	draggable =  draggable || false;
 
 	//init map
   var mapOptions = {
--- a/app/assets/javascripts/maps.js
+++ b/app/assets/javascripts/maps.js
@@ -1,7 +1,10 @@
 var map;
 
-function initMap(x, y, image){
+function initMap(x, y, image, draggable, callback){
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/maps.js
```diff
@@ -1,8 +1,11 @@
 var map;
 
-function initMap(x, y, image){
+function initMap(x, y, image, draggable, callback){
 	image = typeof image !== 'undefined' ? image : currentLocationImage;
 
+	// Draggable defaults to false
+	draggable = (draggable === undefined) ? false : true;
```
Hkly: Are we expecting `draggable` to be a boolean in general? Can this be:
```javascript
draggable = draggable || false;
```

stardust66: Thanks. Changed that. Also I think `maps.js` needs to be cleaned up a bit. There are lots of mixed tabs and spaces on the same line.

mi-wood:
I'm going to merge this per @Hkly review and give it a go in the staging environment. Thanks @stardust66 !

PR #376, opened 2017-09-20
Update to Rails 5.1.4

### LINKED ISSUE #287: Upgrade Rails to 5.1

DeeDeeG:
# Context
- Fixes #287 
- Mostly just upgrades gems, also merges in the new default config files (but I commented out/disabled the changes until we research the new defaults or want to switch over).

# Summary of Changes

- Note: The reason/my rationale for each change is in the commit messages.
- The `rails`, `devise`, `mail_form`, `simple_form`, `activeadmin`, `dotenv-rails`, and `cucumber-rails` gems have been upgraded. These were required in order to move to Rails 5.1.4 without dependency conflicts.
- The `rails app:update` command was run (while logged into the vagrant box), which pulls in the default config files that were added or changed since Rails 5.0. I had to accept or reject these files verbatim, so I accepted them, then reviewed my local `git diff` for each file, and changed things back (manually) until there were no functional changes compared to before this PR. i.e. if there was a functional change I tended to comment it out. If some of our lines were deleted, I added them back in. (This was kind of like resolving merge conflicts in git, but more by-hand.)

    - Much of the config file changes are improved comments or formatting from the upstream Rails community.

# Checklist

- [ ] Added Unit Tests
- [x] CI Passes

# Screenshots

## Before
refuge-production-before

## After
refuge-vagrant-after


DeeDeeG:
By the way, there is one "deprecation warning" for outdated code from before Rails 5.1:

```
DEPRECATION WARNING: ActionView::Template::Handlers::Erubis is deprecated and will be removed
from Rails 5.2. Switch to ActionView::Template::Handlers::ERB::Erubi instead.
(called from <top (required)> at /vagrant/config/application.rb:7)
```

The line that generates this error (line 7 in `config/application.rb`) is:

```
# Require the gems listed in Gemfile, including any gems
# you've limited to :test, :development, or :production.
Bundler.require(*Rails.groups)
```

So this line imports everything in our Gemfile... but we don't even specify Erubis directly in the Gemfile. It does appear in Gemfile.lock... so I assume one of our actual dependencies depends on erubis. I think, logically, we can expect this dependency to be dropped by the time Rails 5.2 comes around, or by definition, that gem will be completely unusable (because its dependency is unusable).

In other words, I think there's nothing obvious we can do about it besides hunt down whatever depends on erubis and update it, or wait for an appropriate update to come out (such an update may not exist quite yet). and I presume this problem will "fix itself" around the time Rails 5.2 has been out for a few months.

config/initializers/new_framework_defaults_5_1.rb
```diff
@@ -0,0 +1,14 @@
+# Be sure to restart your server when you modify this file.
```
DeeDeeG: Thinking of leaving this file in, since we might hypothetically want to change to the new defaults (even though I don't expect us to? Not that I know if they are good/bad defaults, but the research/debugging time it would take to know what they do is probably greater than the reward of changing them.) If we should just get rid of this file, that's okay too.

It looks like I forgot to comment out one of the two new settings also. It doesn't seem to have changed anything for our site, but I can comment it out before this gets merged.

config/initializers/application_controller_renderer.rb
```diff
@@ -0,0 +1,8 @@
+# Be sure to restart your server when you modify this file.
```
DeeDeeG: Looks like this can be deleted rather than commenting out the whole file, if we're not going to use it...

config/cable.yml
```diff
@@ -0,0 +1,10 @@
+development:
```
DeeDeeG: We can delete this, since we don't use Cable IIRC.

bin/yarn
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env ruby
```
DeeDeeG: We can delete this file as well since we don't we don't use yarn (I think? Honestly I don't know but I think yarn is brand new in Rails 5.1 so presumably we're not using it.)

DeeDeeG:
I'm going to delete (and gitignore) some files per your comment above. I can keep iterating until it looks right.

mi-wood:
@DeeDeeG I'm going to go ahead and merge this. Thanks!

PR #268, opened 2016-07-10
Google accessibility audit

clado:
Hi! I ran the [google accessibility audit](https://chrome.google.com/webstore/detail/accessibility-developer-t/fpkknkljclfencbdbgkenhalefipecmb?hl=en) and fixed many of the issues it suggested.
Unfixed issues include the contrast ratios on `.btn-light-purple`, `.greenRating`, `.yellowRating`, and `redRating` elements with white text, and contrast ratios on the google maps suggestions (😕). I increased the font size on the buttons instead of changing the color, because I didn't want to mess with the colour scheme.

PR #386, opened 2017-11-03
Fix text overflow by widening div.navbar-brand (45px -> 16em)

### LINKED ISSUE #379: Limit navbar widh
@DeeDeeG : 

> Also, the "Refuge Restrooms" text in the navbar is flowing over to two lines of text; I think its html "width" should be set (maybe as a style for .navbar-brand), and have it be wider than the text itself. Right now the width seems to be inhereted from the <a> tag, which is 45x45px. It should be about 280px, or 16em.

DeeDeeG:
# Context
- Fixes #379 
- Tidies up a bit after #377 (aka #268)

# Summary of Changes
- Added some width to `.navbar-brand` so that the "Refuge Restrooms" text doesn't overflow to two lines.
- It's only one (short) line of css.

# Checklist

- [ ] Added Unit Tests
- [x] CI Passes

# Screenshots

## Before
screen shot 2017-11-03 at 18 51 42

## After
screen shot 2017-11-03 at 18 52 36

### DIFF
--- a/app/assets/stylesheets/components/nav.scss
+++ b/app/assets/stylesheets/components/nav.scss
@@ -5,6 +5,7 @@
 
   .navbar-brand {
     color: white;
+    width: 16em;
     padding: 13.5px 60px;
     &:hover {
       color: white;


mi-wood:
Oh yes, this was bothering me the other day. Thank you

PR #389, opened 2017-11-18
Use suggested workaround to fix VO for older iOS

### LINKED ISSUE #384: Make App Store icon work with VoiceOver
@DeeDeeG : 

> The App Store download button isn't focusable with VoiceOver, so it would be nice to make that focusable. Not sure how at the moment, but should be doable.

stardust66:
# Context
- Fixes #384.
- As @DeeDeeG said in #384, a [webkit bug](https://bugs.webkit.org/show_bug.cgi?id=145263) in devices with older iOS (before 10?) prevents Safari from focusing on the App Store button in VoiceOver mode. I used the suggested workaround from the discussion on the bug page, which is to include a `role="img"` property on the image tag.

@DeeDeeG Can you help me test this? I've deployed the change to https://jason-refuge-staging.herokuapp.com/

### DIFF
--- a/app/views/pages/index.html.haml
+++ b/app/views/pages/index.html.haml
@@ -19,7 +19,7 @@
 
   .row.even-more-headroom
     .col-xs-12.col-sm-6.col-sm-offset-3
-      = link_to image_tag("app-store.svg"), "https://itunes.apple.com/us/app/refuge-restrooms/id968531953?mt=8"
+      = link_to image_tag("app-store.svg", role: "img"), "https://itunes.apple.com/us/app/refuge-restrooms/id968531953?mt=8"
 
   .row.more-headroom.splash-bottom-padding
     .col-xs-12.col-sm-6.col-sm-offset-3


DeeDeeG:
Sure, I'll give it a go.

(By the way: I've been trying some things on my local machine (with Vagrant) and I haven't had any luck just yet. The behavior is pretty weird with this bug.)

DeeDeeG:
It works. I'm really curious how you got it to work, so I'm definitely going to take a peek at what your changes look like! But yeah, it's working for me, same iPad as before.

DeeDeeG:
It's now possible to focus the button, although it is a little fiddlier getting the focus to stick, or get it to read out the text, compared to other elements on the page. (It keeps acting like your finger has left the element, even when your finger is right on it. That doesn't deselect the element, but it does cancel reading the label out loud. Meaning you can still tap the link, but some users won't know what they're tapping beforehand.) 

"A bit fiddly" is better than "impossible" though.

There are a bunch of other things one is "supposed to do" to have great SVG accessibility, so maybe one of them will make VoiceOver behave less oddly. See: https://www.sitepoint.com/tips-accessible-svg/ . In any case, this is an improvement already for users with old iOS.

PR #391, opened 2017-11-22
Bump bugnag version

MySolace:
Doesn't appear to have any breaking changes, but do please test against my tests. I didn't see anything, neither did my CI.

# Context
- Bumps bugsnag to 6.1.0


tkwidmer:
This seems ok. Thanks for the contribution @MySolace

PR #393, opened 2017-12-04
Proposal for pulling in yelp gender neutral bathroom results

tastycode:
# Context

This is a proposal for adding yelp gender neutral restroom results into the main search. Some issues still need to be worked out, but i didn't want to put more time into it without being sure that this is a idea that would be supported.

- [ ] Add tests
- [ ] What are the side effects of converting `Restroom#near`'s results into an array and adding it to the yelp results
- [ ] Support pagination in the yelp api
- [ ] Consider scraping yelp?
- [ ] Figure out how to distribute the encrypted key for secrets

# Summary of Changes

- /api/search endpoint returns gender neutral restroom results from yelp
- added support for rails 5.1 encrypted secrets

# Checklist

- [ ] Added Unit Tests
- [ ] CI Passes

# Screenshots
```
irb(main):041:0> pp Restroom::Yelp.near(29.970479, -90.0553187).map(&:attributes)
[{"id"=>nil,
  "name"=>"Paladar 511",
  "street"=>"511 Marigny St",
  "city"=>"New Orleans",
  "state"=>"LA",
  "accessible"=>false,
  "unisex"=>false,
  "directions"=>nil,
  "comment"=>nil,
  "latitude"=>29.96318,
  "longitude"=>-90.05476,
  "created_at"=>nil,
  "updated_at"=>nil,
  "downvote"=>0,
  "upvote"=>0,
  "country"=>"US",
  "changing_table"=>false},
 {"id"=>nil,
  "name"=>"New Orleans Cake Café & Bakery",
  "street"=>"2440 Chartres St",
  "city"=>"New Orleans",
  "state"=>"LA",
  "accessible"=>false,
  "unisex"=>false,
  "directions"=>nil,
  "comment"=>nil,
  "latitude"=>29.963751965603,
  "longitude"=>-90.052963781692,
  "created_at"=>nil,
  "updated_at"=>nil,
  "downvote"=>0,
  "upvote"=>0,
  "country"=>"US",
  "changing_table"=>false},```

### CODE REVIEW

File: app/controllers/api/v1/restrooms.rb [hunk 1]
```diff
@@ -46,7 +46,9 @@ class Restrooms < Grape::API
           r = Restroom
           r = r.accessible if params[:ada].present?
           r = r.unisex if params[:unisex]
-          paginate(r.near([params[:lat], params[:lng]], 20, :order => 'distance'))
+          restrooms = r.near([params[:lat], params[:lng]])
+          yelp_restrooms = Restroom::Yelp.near(params[:lat], params[:lng])
+          paginate(restrooms + yelp_restrooms, 20, :order => 'distance'))
```
stardust66: There's an extra `)` here and I think that's what's breaking the tests.

app/controllers/api/v1/restrooms.rb
```diff
@@ -46,7 +46,9 @@ class Restrooms < Grape::API
           r = Restroom
           r = r.accessible if params[:ada].present?
           r = r.unisex if params[:unisex]
-          paginate(r.near([params[:lat], params[:lng]], 20, :order => 'distance'))
+          restrooms = r.near([params[:lat], params[:lng]])
```
mi-wood: I think my preference here would be to just have a `restrooms.near` that composes the active record and yelp `near` requests. That way `near` can be swapped out with any implementation regardless of search backend

app/models/restroom/yelp.rb
```diff
@@ -0,0 +1,33 @@
+require Rails.root.join('lib', 'yelp_client')
+class Restroom::Yelp
+  class << self
+    def near(lat, lng)
+      conn = Faraday.new(:url => "https://api.yelp.com/v3/businesses/search?latitude=#{lat}&longitude=#{lng}&attributes=gender_neutral_restrooms")
+      results = conn.get do |req|
```
mi-wood: Do we need to catch network errors here? We can log an error and return an empty array in that case

mi-wood:
Thanks so much @tastycode for the PR! I left a few comments, but overall this looks good. I'll take a longer look at it when I have some time.

I think the one thing I have to look into is whether or not there's a license issue with this. I'll try to look into that soon and write back into this PR.

config/secrets.yml.enc
```diff
@@ -0,0 +1 @@
+47bUfgjz1qNOUKxDbc8eJz5r/hmQzNem5WlAmA6vb7ys9PL5GmKjhaq3YKRkv2eKeb/zNXY2/ZU6ZnRWKl4ZD4qoAjIkg1K0+SwCAwr6MSxtK5V9wVd65LATPH66AoNtkhJhFfKzpWLpoeeoAUrBJX2alb8AMm5357nXxI4Y38tbeeD2PwUjZ4nRf4LfdhdmFyR5lQx8sofUMEUZLXlMTJVRVI3yZU7uieDlPs/PNBjrqFQ7UjBt1HhIOzSqbBjBYl5oqCBq5B4fgVzY2zLMxaSyun9y2N+TvxrlYk439gdHl/GngK2IHGSUx4PEajBti6mSFBR9vHJryMo2GIqYLpp9rEAGmIdUezlDq1FIpteN4yogkdGz581alYQXFiSmMNuIj3O1L7KCDvm+nBPYwriHXeZJb3KBMsrH2NQ3AIMvyXXdFk7hgtYgDuHOsubDuZdAQKyGFPFeXvGvV5qDG0zcnIDJ7b1mRC8RNFU7TYPb16H2EtUJfO4Zd7SaSbJB5dntPXY92GZocobIB6yLpj47nHsHb5GPL2NOpDK8dDSjxglwpGU8SxKw7QtGYGhNsXb/jAwjDEWRjk+wEBO2D9OsDwdTSHuvoqswYfrTbzXY8Q==--IRLe/45OlnI3gCUs--ogeizYCd9mcc3CiHnGkBtg==
```
tkwidmer: I'm not 100% sure how this works, how do you encrypt /decrypt these. How do we make sure this works in staging and production?

tastycode: We need a way to distribute the encryption key for this file outside of github. `secrets.yml` and `secrets.yml.enc` are a new feature in Rails 5.1. Here's an article providing a bit of background. https://www.engineyard.com/blog/encrypted-rails-secrets-on-rails-5.1

app/models/restroom/yelp.rb
babadabeer: Yes network errors are to be traced for ascertaining that the code problem is not there and only network issues are there.
This happens with high performance media sites like http://vov.media as well that despite everything perfect, network errors at places create function hindrances.

Need to address things duly that we  do not beat behind the bush and jump up to dry he conclusion promptly.

babadabeer:
There are code errors by newly established source. Once source is established it is already researched upon and hence there is least of scope for errors. Some are high tech common sources as used in http://rideonluck.com with all perfection as a few worthy sites use it or worthy businesses use it and has come a long way in time hence possibility of errors minimised.

Thus established codes can be used more or studied si that we learn from previous experiences. Hit and trial will take away life. Hence need to take experiences of hits by others to get perfect.

mi-wood:
Closing since this is a bit stale

Issue #375, opened 2017-09-09
Feature request: autocomplete addresses.


eroncanc:
Hi all,
I found about RR in a NB support fb group. Great work!

Would love to contribute. Is an autocomplete for addresses on app and website desired? Ive used the google maps autocomplete api before and it was pretty nice.

I saw talk about move to OSM, is that happening anytime soon?

DeeDeeG:
> Is an autocomplete for addresses on app and website desired? Ive used the google maps autocomplete api before and it was pretty nice.

That sounds cool. I think as long as it doesn't cause any issues with our API rate limit (or increase fees) that would be a great addition.

Actually, I think we are already supposed to have auto-complete, at least on the splash page. Maybe it's not working because of some HTTPS issues we are having as of late? (#368)

> I saw talk about move to OSM, is that happening anytime soon?

OpenStreetMaps will happen whenever anyone researches how, and proves it can be done! (i.e. implement all the same features we have currently). I personally believe it can be done. If you wanted to look into that, I'd say go for it. I would be interested in making that happen as well and could be of some assistance if you like.

So the only thing holding that up is the R&D time it would take to do it, and we would also want to be sure the implementation is high-quality when it's done.

eroncanc:
thanks for the reply! 
sorry for the late response, i was away but will begin to look into this and will use this issue to implement.

i believe the autocomplete endpoint uses separate rate limit, but will double check.

DeeDeeG:
​Hi again eroncanc,

As I mentioned in my reply above, I thought we might already have auto-complete, but I did agree it wasn't working properly. I can confirm now: We do have it, it just doesn't load right away. It is blocked because of a thing happening with Piwik. If we were able to fix the Piwik issue, we would get autocomplete back.

If you want to know the full details, here is what I found going on with auto-complete at the moment:

We have a script from [Piwik](https://piwik.org/) which is intended to let us know some stats about what users/how many users visit Refuge Restrooms. (Piwik is like open-source Google Analytics.) Our copy is hosted on a server at enspiral.info (as discussed in #200). Our connection to the piwik.js file on that server must be bad, (Or perhaps the file is gone), because we attempt to download the file from there, but the connection times out.

The request for Piwik.js blocks any other requests for javascript files (only for requests that would begin after the piwik request starts) until it is done. 

[...]

PR #395, opened 2017-12-06
Set piwik to use async tracking in production

### LINKED ISSUE #394: Piwik scripts blocks Google Maps Autocomplete script
On page load, the autocomplete feature on the search bar isn't available for about thirty seconds or more because the autocomplete script has been blocked by the synchronous piwik script, and the search bar is blank.

screen shot 2017-12-05 at 9 54 42 pm

screen shot 2017-12-05 at 9 57 44 pm

stardust66:
This prevents the tracking script from blocking the loading of other scripts, such as the Google Maps Autocomplete script. Now, autocomplete is available right when the page loads.

Compare the difference between https://refugerestrooms.org and https://jason-refuge-staging.herokuapp.com (where async tracking has been enabled).

# Context
- Fixes #394.
- Set Piwik to use asynchronous tracking.

# Screenshots

## Before
screen shot 2017-12-05 at 9 57 44 pm

## After
image

### DIFF
--- a/config/piwik.yml
+++ b/config/piwik.yml
@@ -13,7 +13,7 @@ production:
   piwik:
     id_site: 2
     url: analytics.enspiral.info
-    use_async: false
+    use_async: true
     disabled: false
 
 development:


DeeDeeG:
By the way, I looked into this a bit in #375, and it seems like the company that donated a Piwik instance to Refuge isn't serving `piwik.js` anymore on their server, as in we don't really seem to have Piwik tracking anymore.

This PR is a fine workaround and is future-proof, especially if we consider getting Piwik up and running again. It would prepare us to better handle Piwik outages again, going forward. I just wanted to point out that the Piwik tracking isn't going to ever load as things stand at the moment, and we might think about disabling Piwik for now or seeing about spinning up another Piwik instance. I haven't seen/used the data, so someone who has used it might have more to say on that.

I tried your staging branch and it does solve the autocomplete issue. (If this PR is merged we should close #375, because this fixes that issue.)

tkwidmer:
LGTM, thanks @DeeDeeG and @stardust66

PR #412, opened 2018-01-07
update factory girl to factory bot

mi-wood:
# Context
There was a deprecation warning to update the gem to the new name. 

I followed: https://github.com/thoughtbot/factory_bot/blob/v4.9.0/UPGRADE_FROM_FACTORY_GIRL.md


features/step_definitions/restroom_steps.rb
```diff
@@ -1,5 +1,5 @@
 Given(/^a restroom exists in Winnipeg$/) do
-  FactoryGirl.create(:restroom, {name: 'Winnipeg restroom', street: '91 Albert St.', city: 'Winnipeg', state: 'MB', country: 'Canada'}.merge(locations[:Winnipeg]))
+  FactoryBot.create(:restroom, {name: 'Winnipeg restroom', street: '91 Albert St.', city: 'Winnipeg', state: 'MB', country: 'Canada'}.merge(locations[:Winnipeg]))
```
cllns: Since we have `config.include FactoryBot::Syntax::Methods`, we could just do `create(:restroom, ...)`, like is done elsewhere throughout the test suite. I personally prefer not including that, and being explicit instead by including `FactoryBot.` (though in this case, renaming the library results in more diff noise... I don't think they'll rename it again). That's outside the scope of this PR, though.

cllns: Also, [I hate Winnipeg](https://www.youtube.com/watch?v=xLlsjEP7L-k).

Issue #272, opened 2016-10-28
FAQ for Restroom Hosts


ams0026:
Hi All -- 

I'm new to the project and have been looking around for a way I can contribute. One thing I notice is that you don't have a FAQ for business owners/restroom hosts who find themselves on the list. I would like to take a first pass at that. 

Roughly I'd like to tell business owners: 
1.) How they got on this list (Someone used your restroom and thought you should be listed)
2.) Tips on being a good host 
3.) Why safe restrooms are an issue
4.) What reviews mean
5.) How to have your site removed

I would love feedback on whether y'all think this is a need. Thanks!

ams0026:
I'm not sure how to take this issue but I'm going to get started on it.

t-walker:
I think a FAQ page would be great! Maybe there could be a sign that includes some printable materials that or ability to order cooler signage so that they can be proud of their bathroom and make it more obvious.

mi-wood:
I dug up some resources for work a while back, but it looks like I lost them when I switched my slack account :( I'll try to find some again and post them when I get a chance.

teddynicholas:
I'm also new to the project though I've been watching it for a while. I can help with some design aspects of the project if that's necessary. And I'd also love to help with the composition of the FAQ as well.

tkwidmer:
Thats great. We do have some printable signage on the site. As far as I'm concerned we can start writing the FAQ here, and once we have some good answers we can make a PR

tkwidmer:
@ams0026

bbax412:
Hi everyone!  Took a stab at these here at Hack Night -- all edits are most welcome!  Feel free to spin up into a FAQ and adjust as you'd like :)

**1) What is Refuge Restrooms?**
REFUGE is a web application that seeks to provide safe restroom access for transgender, intersex, and gender nonconforming individuals.  Read more about us on our [About page](https://www.refugerestrooms.org/about).

**2) How did my facility get on this list?**
Someone used your restroom and wanted to submit your location as a resource.  All our data is crowd-sourced and user-submitted.

**3) How can I be a good host?**
Thanks for asking!  Often, REFUGE visitors just hope to make a quick stop and be on our way, and like any other visitor, potentially support your business or location while we're there.  Friendly staff and good experiences can go a long way in getting some love for your location in the reviews section of the app (more on that below!)

**4) Why safe restrooms are an issue**
[...]

PR #413, opened 2018-01-07
Spanish translation 

Intimaria:
# Context
- Following from #396 
- Added translation in Spanish. 

# Summary of Changes

- translated en.yml files in config/locale into Spanish (es.yml)
- created separate folders in config/locale: /es_ES for Spanish language files, /en_EN for english english lang files.

PR #415, opened 2018-01-07
update about and business faq

mi-wood:
Turning https://github.com/RefugeRestrooms/refugerestrooms/issues/272 into haml


mi-wood:
I think we could go for increasing the paragraph font size on these too. I think this requires creating a new class for the about page

DeeDeeG:
Just looking over the code changes, and in vagrant, it all looks pretty good.

- minor issue: the business FAQ link isn't working in vagrant. I would suggest a relative path such as `/business_info` rather than `https://refugerestrooms.org/business_info` but just doing that change didn't fix it on my machine. ~~Might have to set something up in `config/routes.rb`?~~
  - ~~Umm, I'm face-palming a bit that I didn't figure this one out quicker...~~
    - Minor typo: The href to `business_info` in `locales/about.en.yml` is just missing an equals sign. e.g.: `href='so-and-so'`
  - I personally believe a relative URL is better so we can test these things in vagrant more easily.
  - applies to link to `about` page from `business_info` page as well

- Just pointing out: this would probably make a merge conflict for #413, so we should probably merge that first for simplicity's sake... Noting however that the new Business FAQ would be untranslated at that point.
  - Maybe we can get a Spanish translation of `business_info` in place in #413 while this PR is still being worked on?
  - Redo this PR on top of develop after #413 is merged for smooth sailing?
    - Alternatively, just open a new PR to translate `business_info`, then merge #413 and hypothetical `business_info` translation PR, re-do this PR on top of develop

- Paragraph text size being bumped up sounds fine once everything else is settled
  - Can help with that if desired

mi-wood:
I tried to do a relative path but with the haml and locale files it was proving to be a pita so I just kinda gave up 😅

DeeDeeG:
Yeah, it's more complicated than it probably should be. (That's computer tech in a nutshell for you?) :P

Here is what I came up with: https://github.com/DeeDeeG/refugerestrooms/commit/6c8d7e14907551c249b8e4de9f26965faca529c7

mi-wood:
That works :) I was trying to use the rails link helpers which made it weird

Intimaria:
OK. will have a look at this later & translate the new information.

DeeDeeG:
I think we can merge this PR if we want to clean up the PRs list a bit.

Bumping paragraph size can be posted as an issue to get to eventually, so we don't forget. Might make for a `good-first-issue`. OR I can try to make time to add such a paragraph size bump to `about` and `business_info` in this PR before merging this. I'd like to see this merged though, since it's so simple and rather thoroughly discussed in #272.

(note: we have 9 open PRs in the queue, all filed within the past month and a half's time! Wow, are we popular! But yeah, it's piling up!)

DeeDeeG:
I added a commit that bumps paragraph size. (Computes to "15.4px" rather than "14px".) Open to comments or suggestions about how to improve the implementation.

DeeDeeG:
Glad this is merged! 

By the way, `business_info.en.yml` ended up outside of the `en` sub-folder of `config/locales`.

Happy to move it over or let you do so, @mi-wood.

Issue #419, opened 2018-01-11
Some pages aren't translation-ready


DeeDeeG:
From #413:

> When looking at @stardust66's PR over at [Intimaria/refugerestrooms:#1](https://github.com/Intimaria/refugerestrooms/pull/1), I noticed that we have some pages that aren't translation-ready.
> 
> We need to update:
> - the "splash" page (homepage),
> - the Contact page,
> - the search results page,
> - the nav bar (at the top)
> - the footer
> - search bar buttons
> - ARIA labels (probably)
> - and if somehow possible, the API documentation page. 
> 
> Making them "translation-ready" would involve taking "hard-coded" sentences and phrases and moving them into a `[something].en.yaml` file, then updating the actual pages to refer to translation strings in place of hard-coded phrases. In other words, expand the existing translation system to also cover the pages I mentioned.

Solving this helps us to progress with #396 and ensure we can more-comprehensively complete #413

DeeDeeG:
The API documentation page still isn't translated, since all the text for that is hard-coded into a javascript file.

And looking forward, if we add new text to the app, said text should use the i18n API.

But this is basically done, so I'm closing this! Woo!

PR #420, opened 2018-01-11
Set locale based on Accept-Language header

stardust66:
This used to be Intimaria#1.

# Context
- A solution to language switching discussed in #413.

# Summary of Changes

- Set locale based on the Accept-Langauge header of the request, which is usually set by the browser. According to the [rails guide on i18n](http://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests)

> The Accept-Language HTTP header indicates the preferred language for request's response. Browsers set this header value based on the user's language preference settings, making it a good first choice when inferring a locale.

Now, the language of the page will change to fit with the preferred language setting of the browser.

## Change Setting
Chrome Settings
image
screen shot 2018-01-10 at 6 28 37 pm

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -19,7 +19,12 @@ class ApplicationController < ActionController::Base
 
   private
     def extract_locale_from_accept_language_header
-      request.env['HTTP_ACCEPT_LANGUAGE'].scan(/^[a-z]{2}/).first
+      accept_language = request.env['HTTP_ACCEPT_LANGUAGE']
+      if accept_language.nil?
+        "en"
+      else
+        request.env['HTTP_ACCEPT_LANGUAGE'].scan(/^[a-z]{2}/).first
+      end
     end
 
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,17 +14,7 @@ class ApplicationController < ActionController::Base
   end
 
    def set_locale
-     I18n.locale = extract_locale_from_accept_language_header
+     I18n.locale = http_accept_language.compatible_language_from(I18n.available_locales)
    end
-
-  private
-    def extract_locale_from_accept_language_header
-      accept_language = request.env['HTTP_ACCEPT_LANGUAGE']
-      if accept_language.nil?
-        "en"
-      else
-        request.env['HTTP_ACCEPT_LANGUAGE'].scan(/^[a-z]{2}/).first
-      end
-    end
 
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,9 +3,23 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
   before_action :mobile_filter_header
+  before_action :set_locale
 
   def mobile_filter_header
     @mobile = true
   end
+
+  def mobile_filter_header
+    @mobile = true
+  end
+
+   def set_locale
+     I18n.locale = extract_locale_from_accept_language_header
+   end
+
+  private
+    def extract_locale_from_accept_language_header
+      request.env['HTTP_ACCEPT_LANGUAGE'].scan(/^[a-z]{2}/).first
+    end
 
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,7 +14,7 @@ class ApplicationController < ActionController::Base
   end
 
    def set_locale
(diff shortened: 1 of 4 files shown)


DeeDeeG:
Thanks for re-doing this.

I did find out while working on #419 that this breaks if the user requests a locale we don't have.

Here is a screenshot when requesting the `ru` (Russian) locale:
about-russian-locale-no-translations

We need some way of defaulting to something we actually have, probably `en`, when we don't have the top user-accepted language. 

Fortunately there are ways of handling this, such as: https://github.com/iain/http_accept_language That is a gem we can add and then have it select the best locale we provide, selected amongst users' requested locales. I am hoping they let us fall back to a default locale when we match none of the user-accepted languages, but it's not super clear so we would have to test. Seems like most browsers might accept `en` as the bottom choice, regardless of locale, but we don't want to break if users manually delete the `en` language in their preferences.

Or if this can be done without installing another gem that's fine by me.

app/controllers/application_controller.rb
```diff
@@ -19,7 +19,12 @@ def set_locale
 
   private
     def extract_locale_from_accept_language_header
-      request.env['HTTP_ACCEPT_LANGUAGE'].scan(/^[a-z]{2}/).first
+      accept_language = request.env['HTTP_ACCEPT_LANGUAGE']
+      if accept_language.nil?
+        "en"
+      else
+        request.env['HTTP_ACCEPT_LANGUAGE'].scan(/^[a-z]{2}/).first
+      end
```
DeeDeeG: This sets the locale to be the top-preferred locale of the user's browser, whether we have that locale translated or not.  Unfortunately, if the top-preferred locale is one we don't have, users see the translation key names in the page, rather than translated strings from an actual locale.

I think we should 
- Go through each of the user's preferred locales and use the first one we have.
- Find a way to revert to `en` when the user's requested locales aren't ones we have.

Can use a gem to handle this, or do it directly in custom Ruby logic.

stardust66:
I used the gem you suggested. With this, the locale gets set to the correct value and when a locale that we don't support gets requested, it defaults to the value of `config.i18n.default_locale` which defaults to `:en`.

DeeDeeG:
Looks good! Going to give it a quick test with the `ru` locale again just to see what happens.

Also, I'm interested in possibly using `language_region_compatible_from(languages)`. If it does what it says it's supposed to, then that would support `zh_ch` and `zh_tw`, whereas `compatible_language_from(languages)` apparently would treat both as the same thing. Who knows, maybe someday we will have multiple regional versions of the same language.

app/controllers/application_controller.rb
```diff
@@ -14,17 +14,7 @@ def mobile_filter_header
   end
 
    def set_locale
-     I18n.locale = extract_locale_from_accept_language_header
+     I18n.locale = http_accept_language.compatible_language_from(I18n.available_locales)
    end
```
DeeDeeG: Consider using `http_accept_language.`**`language_region_compatible_from`**`(I18n.available_locales)`?

According to [this README](https://github.com/iain/http_accept_language/blob/master/README.md#available-methods):

> **language_region_compatible_from(languages)**: Returns the first of the user preferred languages that is also found in available languages. Finds best fit by matching on primary language first and secondarily on region. If no matching region is found, return the first language in the group matching that primary language.
vs:
> **compatible_language_from(languages)**: Returns the first of the user_preferred_languages that is compatible with the available locales. Ignores region.

DeeDeeG:
Okay, it looks good. This is what I get with the `ru` locale preferred, followed by `es-ES` and `es`, and furthermore, I even deleted `en-US` and `en` from my accepted languages. 
ru-locale-set-got-english

The site still gave me the `en` locale (verified in developer tools):
refuge-lang-equals-en

DeeDeeG:
This looks reallly good in further testing. First, I configured the app to believe it supported English, generic Chinese, Chinese specifically for Taiwan [aka traditional Chinese], and Chinese specifically for China [aka simplified Chinese].

` config.i18n.available_locales = [:en, :zh, 'zh-TW', 'zh-CN']`

Here were my results:

| Preferred languages  | What the site gave me | Correct behavior? |
| ------------- | ------------- | ------------- |
|ru, es-ES, es | en | ☑️ |
|zh, ru, es-ES, es | zh | ☑️ |
|zh-CN, zh, ru, es-ES, es | zh-CN | ☑️ |
|zh-TW, zh-CN, zh, ru, es-ES, es | zh-TW | ☑️ |
|zh-CN, zh-TW, zh, ru, es-ES, es | zh-CN | ☑️ |

stardust66:
Thanks for all your testing @DeeDeeG!

Intimaria:
Wow this is fantastic. I admire all of this work!

PR #423, opened 2018-01-14
Implement correct editing behavior

stardust66:
# Context
- Fix issues encountered during #414.
- My changes make it so that editing a bathroom correctly submits to the `create` method of the restroom controller, and redirects to the original bathroom page. For details see my commit messages.

# Screenshots

image

### DIFF
--- a/app/controllers/restrooms_controller.rb
+++ b/app/controllers/restrooms_controller.rb
@@ -86,6 +86,7 @@ private
 
   def permitted_params
     params.require(:restroom).permit(
+      :edit_id,
       :name,
       :street,
       :city,
--- a/app/controllers/restrooms_controller.rb
+++ b/app/controllers/restrooms_controller.rb
@@ -16,7 +16,6 @@ class RestroomsController < ApplicationController
     if params[:edit_id]
       @restroom = find_restroom
       @restroom.edit_id = params[:edit_id]
-      @restroom.id = Restroom.last.id + 1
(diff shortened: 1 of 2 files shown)


stardust66:
The tests won't pass for reasons mentioned in #414.

app/controllers/restrooms_controller.rb
```diff
@@ -16,7 +16,6 @@ def new
     if params[:edit_id]
       @restroom = find_restroom
       @restroom.edit_id = params[:edit_id]
-      @restroom.id = Restroom.last.id + 1
```
mi-wood: I think without this `@restroom.save` will overwrite the existing entry, right? Since `find_restroom` returns the existing ID.

stardust66: Actually, all that the `new` method does is to help render the restroom submission page. The information being set here ends up as part of the form (`/app/views/restrooms/new`). The information in the form gets submitted to the `create` method, which creates a new restroom object from the parameters passed in.
```
@restroom = Restroom.new(permitted_params)
```
Restroom.new generates a new id automatically, so the old restroom isn't being overridden. Also, the old `id` isn't passed as part of `permitted_params` because there's no field for it in the form in the first place. You can test this locally to confirm.

mi-wood: cool, lgtm to me then.

Issue #425, opened 2018-01-15
Problem local testing Vagrant/Rails Server 


Intimaria:
Hi, I have tried to set up a loal testing environment now both with vagrant on a friend's machine and with the rails server (as per @stardust66's instructions in #413 ) and I have gotten the same error messages. I have a feeling it has to do with the way my local git repo is doing but I would really love someone to give me a hand with it as I want to get it going. These are the error messages which are on my machine which are very similar to the ones on my friend's machine running vagrant and VB (which I don't have access to right now)

I was surprised to get the same error messages on both machines

Some are related to `invalid locale data `, which I suspect has to do with recent changes regarding translation which may not be corectly up to date on my local system, 
However, not sure whether that, or the deprecation warnings or, `invalid API key` are the real problem. Or something else? Thank you! 

Here is the full error message on terminal after trying to run rails server: 

```

No valid API key has been set, notifications will not be sent
DEPRECATION WARNING: ActiveSupport.halt_callback_chains_on_return_false= is deprecated and will be removed in Rails 5.2. (called from <top (required)> at /Users/make/Development/git/refugerestrooms/config/initializers/new_framework_defaults.rb:23)
Not notifying due to an invalid api_key
rake aborted!
I18n::InvalidLocaleData: can not load translations from /Users/make/Development/git/refugerestrooms/config/locales/es/business_info.es.yml: #<Psych::SyntaxError: (/Users/make/Development/git/refugerestrooms/config/locales/es/business_info.es.yml): did not find expected key while parsing a block mapping at line 3 column 5>
/Users/make/Development/git/refugerestrooms/app/admin/admin_user.rb:1:in `<top (required)>'
/Users/make/Development/git/refugerestrooms/config/routes.rb:5:in `block in <top (required)>'
/Users/make/Development/git/refugerestrooms/config/routes.rb:1:in `<top (required)>'
/Users/make/Development/git/refugerestrooms/config/environment.rb:5:in `<top (required)>'

Caused by:
Psych::SyntaxError: (/Users/make/Development/git/refugerestrooms/config/locales/es/business_info.es.yml): did not find expected key while parsing a block mapping at line 3 column 5
/Users/make/Development/git/refugerestrooms/app/admin/admin_user.rb:1:in `<top (required)>'
/Users/make/Development/git/refugerestrooms/config/routes.rb:5:in `block in <top (required)>'
[...]

PR #426, opened 2018-01-15
Add versions to all the migrations

stardust66:
In Rails 5, all the migrations are versioned, and inheriting from ActiveRecord::Migration is not supported. In rails 5.1.4, running non-versioned migrations will throw an error. This fixes the issue. The version indicates the rails version that the migration was generated by.

This post: https://blog.bigbinary.com/2016/03/01/migrations-are-versioned-in-rails-5.html. Explains the versioning. Apparently in rails 5.0, not having versions just gave deprecation warnings, but now it throws errors.

# Context
- @Intimaria got this error in #425 when setting up a local development environment.

```
StandardError: An error has occurred, this and all later migrations canceled:

Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class CreateBathrooms < ActiveRecord::Migration[4.2]
/Users/make/Development/git/refugerestrooms/db/migrate/20131128044650_create_bathrooms.rb:1:in `<top (required)>'

Caused by:
StandardError: Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class CreateBathrooms < ActiveRecord::Migration[4.2]
/Users/make/Development/git/refugerestrooms/db/migrate/20131128044650_create_bathrooms.rb:1:in `<top (required)>'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)
```


DeeDeeG:
I see this is an issue, but I have trouble understanding why this happens. If you don't mind, a couple quick questions:

1. Is this something that happens only outside of a Vagrant environment?

2. How might someone reproduce these errors?

stardust66:
My guess is that people haven't been running the migrations after we moved to rails 5, since they'd already set up their database, so nobody's run into this issue. Only people who are setting up their databases for the first time or making changes to the database will encounter this issue. [The migration](https://github.com/RefugeRestrooms/refugerestrooms/pull/414/files#diff-81b6b5590da864b0d34952f53d0aff7aR1) generated by the new version of rails in #414 comes with the version number.

DeeDeeG:
Perhaps someone who deals with the database more will understand the significance of this more than I do.

I notice we have had various Rails versions between 4.0 to 5.0, and then we bumped straight to 5.1.4... These migrations have been added at various times when there were various Rails versions in the Gemfile. Does it matter if the Rails version we add in this PR exactly matches the Rails version we were actually using? Or since migration versioning is introduced in Rails 5, is it a moot point before Rails 5.0?

That's about as deep as I am able to go in reviewing this.

stardust66:
According to the post linked in the first comment, Rails handles migrations differently starting from version 5, so all the versions before should have the same behavior. Also, you can't actually set the migration version to anything before 4.2, or rails will throw this error when migrating:
```
ArgumentError: Unknown migration version "4.1"; expected one of "4.2", "5.0", "5.1"
```

DeeDeeG:
I just tried this in Vagrant, by deleting my virtual machine and starting over (`vagrant up`) from scratch.

For whatever reason, given the way the [script](https://github.com/RefugeRestrooms/refugerestrooms/blob/develop/setup/setup_vagrant.sh#L97) [sets up](https://github.com/RefugeRestrooms/refugerestrooms/blob/develop/db/seeds.rb) the environment/db, none of these errors happen.

I still get the impression this PR is "the right thing to do," in some sense, but I'm not sure in what well-defined context it's needed, either.

I am curious (because I don't know what all this means): in [this comment](https://github.com/RefugeRestrooms/refugerestrooms/issues/425#issuecomment-357801083) of #435, you suggested running `rails db:migrate`.

(Is this a useful command to run? Should we expect to run this command often?)

As far as I can tell, that command is the only one that has errors (fixed in this PR), whereas the setup script only needs/uses `rails db:setup`.

stardust66:
I just figured this out. The Vagrant script runs [`rails db:setup`](https://github.com/RefugeRestrooms/refugerestrooms/blob/develop/setup/setup_vagrant.sh#L99), which runs `db:create`, `db:schema:load`, and `db:seed`, according to [this stack overflow post](https://stackoverflow.com/questions/10301794/difference-between-rake-dbmigrate-dbreset-and-dbschemaload). `db:schema:load` changes the database structure based on what's in `schema.rb`, which is generated when migrations are run for the first time. 

`db:schema:load` wipes all data in the database, while `db:migrate` gives versions to changes to the database structure and lets you roll back to a specific version while keeping the data intact. So, `db:setup` is a better command to run when you're first setting up the database, since it's one command that does everything you need, but `db:migrate` is for when you want to change the database structure while keeping the data.

In short, #425 I should have recommended `rails db:setup` because @Intimaria was setting up the environment for the first time. If everyone uses that command, technically nobody would run into any errors because the old migrations wouldn't have to be run at all, because the database would be built based on `schema.rb`, and any new migrations (like the one made by @mi-wood for #414) would already have version numbers on them. However, I think we should merge this just in case people are setting up their environment without vagrant and are running `db:migrate` instead of `db:setup`.

Intimaria:
Also, and I don’t know if this makes any difference, I am running the app locally using rails directly from terminal and without Vagrant. I had gotten the same errors in both initially. This solution fixed it on my machine where I was running from terminal (however, I haven’t had access to vagrant again since to check this stuff there). 
Would I meed to run the 

‘db:setup’

instead of 

‘db:migrate’

on terminal as well?

DeeDeeG:
I don't know much about this stuff, but I do know how to check if the db is set up properly. Here's how to check:

- Use the search bar on the Refuge Restrooms page that you are running locally, and search for "San Francisco" (or even "One Embarcadero Center, San Francisco, CA")
- You should see a bunch of restroom results
- And when you click on one of the results, the restroom info should look normal. (It should look similar to how it looks at [refugerestrooms.org](https://refugerestrooms.org/restrooms/33178))

Basically, if the restroom entries are working, everything is fine.

(Correct me if I'm wrong @stardust66)

If you want to run `rails db:setup` again, I don't think it hurts anything to do so. If it errors out, but you have restroom results in search, I personally wouldn't worry about it.

stardust66:
Yes @DeeDeeG you are right. @Intimaria You might not have run `rails db:seed`, which would make your database not have the initial data. That command is run when you run `rails db:setup`, but you can also do it separately.

DeeDeeG:
If there *are* errors, I expect they would be "invalid locale data" again, and would have to do with `config/locales/es/landing.es.yml`, which is no longer used and can (probably should) be deleted.

Issue #411, opened 2018-01-07
Vagrant up timeout


padalah:
I am newcomer to OSS and I really want to contribute this project. When I run the command "vagrant up", I keep getting a time-out error.

mi-wood:
Hi @padalah and welcome! Could you please past your error and let us know what operating system you're using?

DeeDeeG:
Hi @padalah

If the error looks something like this: 

```
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: vagrant
    default: SSH auth method: private key
    default: Warning: Remote connection disconnect. Retrying...
    default: Warning: Remote connection disconnect. Retrying...
    default: Warning: Remote connection disconnect. Retrying...
    default: Warning: Remote connection disconnect. Retrying...
    default: Warning: Remote connection disconnect. Retrying...
```

That is normal. You only need to wait a minute or two, and it will connect successfully.

Then you should see something like this:

```
==> default: Machine booted and ready!
```

I also found some threads on stackoverflow.com where people discussed similar-sounding issues:
 - https://stackoverflow.com/questions/22575261/vagrant-stuck-connection-timeout-retrying
 - https://stackoverflow.com/questions/23690124/vagrant-up-timeout

I hope one of those links can help.

(And to second @mi-wood, welcome!)

(Also to second @mi-wood, we would possibly be able to help better if you can copy and paste the error you are seeing. Good luck!)

padalah:
My operating system is Windows 10. Here is the error I am encountering:
<img width="454" alt="refrest1" src="https://user-images.githubusercontent.com/19918773/34707287-88aa8c6e-f4c1-11e7-9ca0-043664c631cc.PNG">
<img width="381" alt="refrest2" src="https://user-images.githubusercontent.com/19918773/34707290-8e81f690-f4c1-11e7-83da-b285621e8a38.PNG">

DeeDeeG:
I would run `vagrant destroy`, delete any `.vagrant` folder within your RefugeRestrooms folder, then try `vagrant up` again.

I think there was just a mis-configuration when bringing up the virtual machine the first time, and starting over might clear it up.

If that still doesn't work, maybe your version of Vagrant or VirtualBox is out-of-date?

I also found a blog post about dealing with this problem, if none of that helps: https://tomjn.com/2016/08/03/vagrant-provision-timeouts/

PR #435, opened 2018-01-29
Docker

mi-wood:
I'll hunt down any vagrant issues, as this will close them


mi-wood:
Closes https://github.com/RefugeRestrooms/refugerestrooms/issues/411 https://github.com and /RefugeRestrooms/refugerestrooms/issues/336

stardust66:
Running `docker-compose down` wipes the development database, since it's a temporary change made by the `docker-compose run` command. Is there a way to make this permanent? I tried putting this in the `Dockerfile`, which is only run during the building phase, but it didn't work because `db` hasn't been built yet. Is there a way to make docker-compose run something only at build time?

mi-wood:
I'm crying, this is so much easier

stardust66:
Hey, I think we can avoid this problem altogether by not running `docker-compose down`, since pressing `ctrl-c` already stops the services. `docker-compose down` removes the containers, which I don't think we need to do.

mi-wood:
`ctl-c` doesn't kill the rails process itself, so next time you run `docker-compose up` there might be a phantom process that doesn't allow it to run

CONTRIBUTING.md
```diff
@@ -3,35 +3,39 @@
...
 
-### 5 Optional tasks:
-run <code>rake db:fix_accents</code> to clean up encoding problems in the safe2pee data. (Use <code>rake db:fix_accents[dry_run]</code> to preview the changes.)
+### 5 Run the Tests
+```
+docker-compose run -e "RAILS_ENV=test" web rake db:test:prepare spec cucumber
+```
 
-### 6 Assets
+### 6 Optional tasks:
```
mi-wood: @tkwidmer do we need this anymore? I feel if we do, it might not belong in this doc

tkwidmer: I'm not sure if we still need this. The seed has been reduced to a small sample set rather than a larger set that it used to be. So we may not need this.

Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ruby:2.3.1
+RUN apt-get update -qq && apt-get install -y build-essential libpq-dev nodejs
+ENV PHANTOM_JS=2.1.1
+RUN  apt-get update && \
+  apt-get install build-essential chrpath libssl-dev libxft-dev -y && \
+  apt-get install libfreetype6 libfreetype6-dev -y && \
+  apt-get install libfontconfig1 libfontconfig1-dev -y && \
+  cd ~ && \
+  export PHANTOM_JS="phantomjs-2.1.1-linux-x86_64" && \
```
mi-wood: just reminding myself this should be $PANTOM_JS not 2.1.1

CONTRIBUTING.md
mi-wood: I'm going to delete since in my past 3.5 years of working on this I've never done this. It might just be confusing to a newcomer

stardust66:
I think the rails server leaves a `server.pid` file inside `tmp/pids/` which prevents a new server from starting up.
image

My workaround is to mount an empty volume at `/refugerestrooms/tmp/`. In `docker-compose.yml`
```
services:
  web:
    volumes:
      - .:/refugerestrooms
      - /refugerestrooms/tmp
```
This makes the tmp folder in the container initially empty. Thus, no `server.pid` file to prevent the server from starting. This also has the added benefit of persisting development data (any bathrooms you add to the database for testing will stay).

@mi-wood Do you see any downsides to this?

mi-wood:
@stardust66 I'd probably rather just seed the DB every time than implement a workaround and have people use `ctl-c` since it goes against the docs. The seeding only takes a few seconds. If it becomes an issue, we can always change it.

stardust66:
Okay. That makes sense.

mi-wood:
Going to go ahead and merge this, so I don't have to cherry pick around other branches to use it

tkwidmer:
👍

PR #441, opened 2018-01-30
update piwik url

mi-wood:
Clean version of https://github.com/RefugeRestrooms/refugerestrooms/pull/440


app/views/layouts/_piwik_tag.html.haml
```diff
@@ -0,0 +1,14 @@
+<!-- Matomo -->
+:javascript
```
mi-wood: trying to decide if I should guard this based on env. Probably...

mi-wood:
Going to add this to the next release

Issue #347, opened 2017-05-21
Switch production web server from WEBrick to Puma or similar


DeeDeeG:
Hi folks,

I was reading on the Heroku site, and I saw that the recommended web server software to run is Puma, according to Heroku. https://devcenter.heroku.com/articles/ruby-default-web-server

(There is also a web server called "Unicorn". Who knew?)

The way this is described in the guide, it's not actually too hard to switch. (Mostly just change our repo's Procfile (which is one line long) and optionally make a config file that can pretty much be copy-pasted from their guide if we want to tweak/explicitly set certain settings.)

Guide: https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server

The only thing I noticed that might be a tricky: We should set our min-threads and max-threads to 1, just in case our code isn't "thread safe," or do some debugging on a staging instance and just make sure nothing breaks with more than one thread enabled before we deploy it to production.

The benefits should be speed, mostly.

DeeDeeG:
I could do this. Testing would involve setting up a Heroku instance, which I've never done before, so it would possibly take me longer to test it than anyone who has already deployed to Heroku, but I think it would be surmountable.

PR #442, opened 2018-01-30
move to puma

mi-wood:
# Context
- Fixes https://github.com/RefugeRestrooms/refugerestrooms/issues/347
- Move to puma as recommended by Heroku

# Summary of Changes

- adds puma gem and changes procfile
- thread env variable has been updated on heroku

### DIFF
--- a/app.json
+++ b/app.json
@@ -14,18 +14,6 @@
     "EMAIL_PASSWORD": {
       "required": true
     },
-    "HEROKU_POSTGRESQL_AQUA_URL": {
-      "required": true
-    },
-    "HEROKU_POSTGRESQL_CYAN_URL": {
-      "required": true
-    },
-    "HEROKU_POSTGRESQL_NAVY_URL": {
-      "required": true
-    },
-    "HEROKU_POSTGRESQL_PURPLE_URL": {
-      "required": true
-    },
     "LANG": {
       "required": true
     },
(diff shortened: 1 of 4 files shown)


tkwidmer:
This looks good to me, have we deployed in a test environment to make sure it works?

app.json
```diff
@@ -14,18 +14,6 @@
     "EMAIL_PASSWORD": {
       "required": true
     },
-    "HEROKU_POSTGRESQL_AQUA_URL": {
```
mi-wood: @tkwidmer I imagine these are old DB names?

DeeDeeG:
> thread env variable has been updated on heroku

This could be set to "one," to make sure we are de-facto "thread safe," or this could be two or more, for speed.

Which way are we thinking of going with this? Would anyone be interested in trying multiple threads anyway, just to see if everything works? Or play it more safely by sticking with one thread?

 (Issues, if any, would likely be intermittent race conditions, and therefore harder to diagnose than some issues are. Also, since most of our Ruby code is third-party via gems, fixing the issues could involve working upstream with those gem maintainers. But if there are no such issues, it's all moot and we can use however many threads we want.)

mi-wood:
@DeeDeeG I set it to 1 and figure that should be fine. I don't think using one thread is going to be an issue performance wise.

I also created https://github.com/RefugeRestrooms/refugerestrooms/issues/445 to evaluate this for the future.

PR #446, opened 2018-02-06
add default en form fields

mi-wood:
# Context
- Adds defaults for form translations


mi-wood:
@DeeDeeG that i18n debug gem is soooo helpful

PR #448, opened 2018-02-09
.gitattributes: Unix-ify line-endings for Docker

DeeDeeG:
# Context
- Fixes an issue when using Docker on Windows.
  - Docker doesn't undestand the `entry` script if the file is checked out with Windows-style line endings (`CR-LF`), which is how Git checks the file out by default on Windows. Setup errors out. Changing this fixes the error.

# Summary of Changes
- Uses a `.gitattributes` file to configure the `setup/entry` file to always have Unix-style line-endings (`LF`).

# Checklist

- [ ] Added Unit Tests
- [x] CI Passes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+setup/entry text eol=lf


DeeDeeG:
Thanks for the review!

PR #457, opened 2018-02-17
convert cucumber to capybara

mi-wood:
# Context
- Move off of cucumber to decrease time to write controller test


mi-wood:
I'm having trouble figuring out why adding poltergeist is causing these timeout, since the same tests worked with cucumber. I imagine it's some default cucumber setting being applied that is not with capybara.

If someone else could take a look, that would be a huge help. maybe @stardust66 or @DeeDeeG .

mi-wood:
I didn't realize that this is also failing in `develop`, which means something has happened since our last merge.

DeeDeeG:
Oddly enough, this: https://github.com/RefugeRestrooms/refugerestrooms/pull/448 is the one PR related to docker that is at about the right spot in the commit history. Maybe try removing the `.gitattributes` file for this branch, and push to this PR so Travis retries?

DeeDeeG:
By the  way, I'm getting a lot of useful info to track down this problem by viewing the history of all PRs, including closed ones. That shows about when this started happening (where red xes start showing up.)

https://github.com/RefugeRestrooms/refugerestrooms/pulls?utf8=✓&q=is%3Apr

mi-wood:
@DeeDeeG it times out with or without that file. also, the most recent develop merged would have failed if that had something to do with it. 

I'm thinking it's a) something new unrelated to a change (external api calls or something) or b) something flaky that's only happening some time that was introduced recently

I'm leaning towards a, since I've yet to get the tests to pass and they seem to be failing consistently in the same spot

DeeDeeG:
I'd like to set up a PR or two exclusively for some Travis debugging. Will do soon.

DeeDeeG:
Would folks be interested in setting up Travis _not_ to use Docker, just to keep the Travis setup simpler, more low-level, hopefully less moving parts and less breakable?

DeeDeeG:
> I'm thinking it's a) something new unrelated to a change (external api calls or something) . . .

After trying some things, I have to agree with this. I do not think this is a problem we caused ourselves in this repo. It looks more like one of our dependencies has a bug.

All my recent PRs tested various small changes we could make. All of said PRs still had the timeouts. So that seems to rule out a lot of the possible causes within our repo.

mi-wood:
I cannot figure out what might be wrong here. I tried a few more things:
- running outside of docker
- removing the piwik url (since that was a recent change)

Still having no luck tracking this down. It's so bizarre that a change in our code didn't trigger this

DeeDeeG:
For what it's worth (not a lot in this case?) They seem to be baffled
by some similar things upstream.

https://github.com/teampoltergeist/poltergeist/issues/904
https://github.com/teampoltergeist/poltergeist/issues/375

We can possibly fix this by moving to a new headless browser. See this
for a guide:
https://engineering.upgrad.com/turn-your-website-testing-painless-with-chrome-and-firefox-headless-92b6f023d375

(Steps 4 and 5 are a very concise couple-of-sentences-worth how-to,
the rest of the article gives good context.)

mi-wood:
I'm going to deploy this, since the current issue with the maps API is causing browsers to crash :(

Issue #458, opened 2018-02-18
Allow Docker to Launch Browser for Poltergeist


mi-wood:
This will allow debugging browser tests. My thoughts are we need to
- Install chromium
- Use this to get an idea of running a desktop app in docker: https://blog.jessfraz.com/post/docker-containers-on-the-desktop/
- Figure out X11 path through

PR #459, opened 2018-02-18
Add chromium and set up X11 socket

stardust66:
This didn't work on my machine, but it should as long the X11 server on the host is allowing connections from the docker container. This involves using 'xhost' to specify permissions.

You'll need XQuartz if you're on a Mac, or Cygwin/X if you're on Windows.

# Context
- #458

# Summary of Changes

- Attach host X11 socket (`/tmp/.X11-unix`) to container.
- Expose environmental variable `DISPLAY` to container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,4 +16,5 @@ WORKDIR /refugerestrooms
 COPY Gemfile /refugerestrooms/Gemfile
 COPY Gemfile.lock /refugerestrooms/Gemfile.lock
 RUN bundle install
+RUN apt-get install -y chromium
 COPY . /refugerestrooms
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -16,4 +16,5 @@ WORKDIR /refugerestrooms
 COPY Gemfile /refugerestrooms/Gemfile
 COPY Gemfile.lock /refugerestrooms/Gemfile.lock
 RUN bundle install
+RUN apt-get install -y chromium
```
stardust66: I put this at the end instead of with all the `apt-get` statements so Docker can use its cache for everything that came before.

mi-wood:
I tried something similar and am still getting "Cannot open display" :/

stardust66:
Yeah that's what I've been getting. Did you change the permissions with ‘xhost’?

PR #487, opened 2018-06-13
User edit proposals - rebased on develop

stardust66:
# Context
- Continuation of #414, mostly @mi-wood's work.

# Summary of Changes

From #414:
> 1. Add a column for edit_id and approved. Only listings with approved = true will show up, and the results will be grouped by edit_id using the most recent id
> 2. Allow users to propose edits that will add a new row with edit_id equal to the original listing id
> 3. The admin just has to set approved = true for approved edits in active admin.
4. Add migration to update the edit-ids of existing restrooms to their ids.

This migration is fine if no edits were added prior to running it (and rails shouldn't let that happen). Someone more familiar with migrations should take a look though.

# Checklist

- [x] Added Unit Tests
- [ ] CI Passes

# Screenshots of Editing Process


DeeDeeG:
This looks really cool!

I wouldn't know how to test that this works. But this has been a big user-requested feature for some time, so it would be pretty awesome to be able to do this.

It certainly looks functional in your screenshots, too. 👍

mi-wood:
I'd like to try to push this through soon, since we're so behind on edits. Looks like code climate is yelling, but other than that, this seems good to me.

stardust66:
I'm not sure why Code Climate doesn't like it. The page has nothing and pressing analyze hasn't done anything yet. Maybe we can trigger a rerun with another commit?
image

DeeDeeG:
The "Analyze" button took me to a totally empty/blank page, if I wasn't logged in, but when I was logged in to CodeClimate with my GitHub credentials and clicked it again, it told me it had "requested a new analysis":

Code Climate: "We have requested a new analysis."

Hopefully that makes Code Climate happy. . .

DeeDeeG:
Otherwise, I do think adding another commit could cause Code Climate to refresh.

(In situations like these I have tended to do "git commit --amend" and "git push -f", rather than create actual new commits, since dummy commits can look odd in the commit history.)

DeeDeeG:
Okay, you can disregard my previous two comments.

I think this is just because your branch doesn't have the code climate stuff in it: https://github.com/stardust66/refugerestrooms/blob/user-edits/.travis.yml

Ways to fix:

- I suppose we could merge develop into this branch, and that should take care of it. 
- Or one could rebase on develop _again_ and force push to this branch.
- Or manually set up code climate again (exactly the same way to avoid merge conflicts) on your branch. 

Any of those should work, probably.

DeeDeeG:
Okay, looks like this has been rebased now. Thanks @stardust66.

CodeClimate has analyzed this, and says the `create` method in [`restrooms_controller.rb`](https://github.com/RefugeRestrooms/refugerestrooms/pull/487/files#diff-1c9d19bd5b15ef3f0e92bd93aad415b3R29) should be simpler / easier to read.

So, looking into this, we can either try to refactor / break this up and somehow make it simpler, or rather decide it's okay as-is. I don't read/edit the ruby code very much, so other folks' input would be appreciated on this one. Otherwise, there are no remaining issues with CodeClimate or Travis.

DeeDeeG:
The merge conflict here is really small, by the way:

```diff
-ActiveRecord::Schema.define(version: 2015_10_18_191859) do
+ActiveRecord::Schema.define(version: 20180107182858) do
```

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180613231032) do
+ActiveRecord::Schema.define(version: 2015_10_18_191859) do
```
DeeDeeG: Travis apparently reads this older date and thinks we have "pending migrations' and fails.

I recommend just using the same date (2018 06 13, 23:10:32) as before, with or without the underscores. (I don't get what this all means, to tell the truth, but I think it would make Travis pass the tests again.)

PR #498, opened 2018-09-24
Gemfile[.lock]: Update out-of-date gems

DeeDeeG:
## Context
- Bumps a few Gems in `Gemfile.lock`, and one in `Gemfile`

There was an issue with more recent versions of grape (or grape- we should look out for: #210 AKA #203. Maybe check the API demo page and double-check that works.

## Summary of Changes

- Updated `ffi` with command `bundle update ffi`
- Updated `sprockets` with command `bundle update sprockets`
- Manually updated `grape` in `Gemfile`, then ran `bundle update grape`

## Checklist

- [ ] Added Unit Tests
- [ ] CI Passes

### Before
_(no screenshots at the moment)_
### After
_(no screenshots at the moment)_


DeeDeeG:
I did experience some problems with the API page.

It seems like we need to update `grape-swagger`, based on this compatibility table: https://github.com/ruby-grape/grape-swagger#compatibility- 

(I mean that we pretty definitely want the latest `grape` gem, which is 1.1.0, and I think this is incompatible with the older `grape-swagger` we have at the moment, 0.7.2.)

And it then looks like we need to do some configuration changes, based on this UPGRADING instructions file for `grape-swagger`: https://github.com/ruby-grape/grape-swagger/blob/master/UPGRADING.md

DeeDeeG:
We can look back into this after #505 and/or #506.

DeeDeeG:
This is completely taken care of by #505 and #506.

We can't upgrade `grape` without also upgrading `swagger-ui`, so #506 is the way to go to finish this up.

Closing in favor of #505 and #506.

PR #499, opened 2018-09-30
New generic feature proposal issue template

DeeDeeG:
Generic issue template for any proposed change. Has sections to add context and guidance for how to implement the proposal. Designed to make issues clear and easy to understand/work on.

### DIFF
--- a/.github/ISSUE_TEMPLATE/generic-proposal-issue-template.md
+++ b/.github/ISSUE_TEMPLATE/generic-proposal-issue-template.md
@@ -1,0 +1,43 @@
+---
+name: Generic Proposal Issue Template
+about: Generic issue template for any proposed change. Has sections to add context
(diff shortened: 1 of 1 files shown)


DeeDeeG:
I dunno 🤷. 

Seemed like a useful Issue Template for making Hacktoberfest issues.

This template asks for / has headings for a lot of context that might be useful for someone new to the project. So I think it would help us make issues that contributors have a good experience with.

Open to comments and feedback. Also, if no-one plans to use this, I can keep it as a `.md` file on my computer somewhere; it needn't become an official part of the project in that case. I'm not sure how often this would come in handy, but if we do a lot of issues for Hacktoberfest, I hope that it would.

tkwidmer:
I think this is great at least as a template. I'm in favor of templates.

PR #501, opened 2018-10-02
Update font stack

### LINKED ISSUE #494: [CSS] Add additional good-quality fonts to the web app
## Scope / difficulty

We just need to change one line in our [`_variables.scss`](https://github.com/RefugeRestrooms/refugerestrooms/blob/ee25f6af625d06269c77377a3ea594308a644de4/app/assets/stylesheets/sass/_variables.scss#L3) file.

## Impact

This would make the app look a bit nicer on most operating systems and browsers.

## Details

Basically, we already have "Ubuntu" as a really nice font, and then we currently fall back to variations on "Helvetica" or else we fall back to [insert sans-serif font likely to look like Helvetica here].

I think we should add some other nice fonts that people probably have on their computers.

### Rationale

I thought to do this, because there are actually some great "basic sans-serif" fonts out there that people already have, which might be even better than Helvetica / Arial and equivalent.

### Proposal

What I have in mind is `Segoe UI` (almost every Windows user has this, since Vista and onward), [`-apple-system`](https://webkit.org/blog/3709/using-the-system-font-in-web-content/)[[1](https://stackoverflow.com/questions/32660748/how-to-use-apples-new-san-francisco-font-on-a-webpage#answer-32660790)] (recent macOS and iOS versions will use `San Francisco`; all the way back to the iPad 2 and similarly old iPhones support this, and I believe many Mac owners are on the latest macOS, which will also use `San Francisco`.) And we can consider some common Android fonts, such as `Roboto` and even `Droid Sans` for older Android phones.)

See this list of common "system" fonts: https://css-tricks.com/snippets/css/system-font-stack/

## How to actually do this:

One would just need to update this line of our `app/assets/stylesheets/sass/_variables.scss` file: [link](https://github.com/RefugeRestrooms/refugerestrooms/blob/ee25f6af625d06269c77377a3ea594308a644de4/app/assets/stylesheets/sass/_variables.scss#L3)

Before merging any changes, it would be great to have someone test this and ensure it works, and take some time to consider the style pros/cons of these fonts.

lucas-carl:
# Context
- Fixes #494 
- Use better fallback fonts that users have on their systems

# Summary of Changes

- Updated `$font-stack` SASS variable

### DIFF
--- a/app/assets/stylesheets/sass/_variables.scss
+++ b/app/assets/stylesheets/sass/_variables.scss
@@ -1,6 +1,6 @@
 /*All From Susy File*/
 
-$font-stack: "Ubuntu", "Helvetica Neue", Arial, Gotham, "Helvetica", sans-serif;
+$font-stack: Ubuntu, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Cantarell, "Helvetica Neue", sans-serif;
 $primary-color: #333333;
 $purple: #41326B;
 $light-purple: #8377af;


DeeDeeG:
Will wait for some comments from other maintainers, and if this pull request gets to be 48 hours old without further comments I plan to merge.

Thanks again @lucas-carl, and happy Hacktoberfest. 🎃 💻 🍂 ✨

DeeDeeG:
Here are some screenshots on iPad 2 running iOS 9.3.5:

## Before

Refuge web app using Helvetica

(Uses Helvetica or Helvetica Neue, I'm not 100% sure)

## After

Refuge web app using San Francisco

(Uses San Francisco)

DeeDeeG:
Here are some screenshots on a PC running Ubuntu 18.04, with various fonts installed:

## Before

Refuge web app using the Ubuntu font

(Uses Ubuntu)

## After

Refuge web app using the Roboto font

(Uses Roboto)

Refuge web app using the Oxygen Sans font

(Uses Oxygen Sans)

Refuge web app using the Cantarell font

(Uses Cantarell)

PR #504, opened 2018-10-04
Replace ruby base-image in Dockerfile with slimmer version

darpr:
# Context
- Fixes #[issueNumber]
- Replaced ruby base-image (which was heavy at 941 MB) with a slimmer base-image (just 254 MB)

# Summary of Changes

- Used a slimmer base image which has a footprint of only 254 MB (faster deploy times, across cloud)
- Since slimmer base image missed some packages required during "docker build", added lines to Dockerfile to install basic binaries

# Checklist

- [ ] Added Unit Tests
- [x] CI Passes

Tests passed:
```
$ docker-compose run -e "RAILS_ENV=test" web rake db:test:prepare spec
Creating network "refugerestrooms_default" with the default driver
Creating refugerestrooms_db_1 ... done
Created database ‘bathrooms_test'
..
..
..
Randomized with seed 54156
.....................................Puma starting in single mode...
* Version 3.11.2 (ruby 2.3.7-p456), codename: Love Song
* Min threads: 0, max threads: 1
* Environment: test
* Listening on tcp://0.0.0.0:39645
Use Ctrl-C to stop
...................

Finished in 23.82 seconds (files took 8.49 seconds to load)
56 examples, 0 failures

Randomized with seed 54156

Coverage report generated for RSpec to /refugerestrooms/coverage. 541 / 594 LOC (91.08%) covered.
```

# Screenshots
Could launch application after the changes:
image

## Before
With the previous base-image, host machine had a 941 MB Ruby base image.

## After
With these changes, the Ruby base image footprint on host is just 254 MB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,10 @@
-FROM ruby:2.3.7
+FROM ruby:2.3.7-slim
 ENV PHANTOM_JS=2.1.1
+
+# Add basic binaries and clean up the apt cache
+RUN apt-get update \
+  && apt-get install -y bzip2 curl gnupg wget \
+  && rm -rf /var/lib/apt/lists/*
 
 # Add the apt repository for yarn
 RUN curl -sS http://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,9 @@
 FROM ruby:2.3.7-slim
 ENV PHANTOM_JS=2.1.1
 
-# Add basic binaries and clean up the apt cache
+# Add basic binaries
 RUN apt-get update \
-  && apt-get install -y bzip2 curl gnupg wget \
-  && rm -rf /var/lib/apt/lists/*
+  && apt-get install -y bzip2 curl gnupg wget
 
 # Add the apt repository for yarn
 RUN curl -sS http://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
@@ -24,6 +23,10 @@ RUN apt-get install build-essential chrpath libssl-dev libxft-dev -y && \
   mv $PHANTOM_JS /usr/local/share && \
   ln -sf /usr/local/share/$PHANTOM_JS/bin/phantomjs /usr/local/bin && \
   apt-get install -y yarn
+
+# Clean up the apt cache
+RUN rm -rf /var/lib/apt/lists/*
+
 RUN mkdir /refugerestrooms
 WORKDIR /refugerestrooms
 


DeeDeeG:
This looks cool, thanks for the pull request!

I am in favor of merging, and we may have another maintainer get a chance to look at this over the weekend. Will wait until then so other maintainers can discuss and give their thoughts.

In any case, happy Hacktoberfest!

darpr:
Thanks for the feedback, @DeeDeeG! Appreciate the work that you guys are doing through this project.

tkwidmer:
This looks fine to me. @mi-wood do you want to weigh in as the only person with more Ops experience?

Dockerfile
```diff
@@ -1,6 +1,11 @@
-FROM ruby:2.3.7
+FROM ruby:2.3.7-slim
 ENV PHANTOM_JS=2.1.1
 
+# Add basic binaries and clean up the apt cache
+RUN apt-get update \
+  && apt-get install -y bzip2 curl gnupg wget \
+  && rm -rf /var/lib/apt/lists/*
```
DeeDeeG: I think `rm -rf /var/lib/apt/lists*` is intended to clean up some disk space and make the Docker container stay small on developers' hard drives.

But we do more `apt-get update` later in the`Dockerfile`, so I don't think this would help much. We will download the contents of this folder again, so this ends up requiring more data to be downloaded, I think. I think the hard drive space being a tad bigger is better than having to download more MB of data from the network.

And we would want to be more consistent with removing this folder's contents after every `apt-get update` to really save hard-drive space.

See: _"what is the purpose of the files in this directory: /var/lib/apt/lists?"_ https://serverfault.com/questions/449726/what-is-the-purpose-of-the-files-in-this-directory-var-lib-apt-lists

DeeDeeG:
Just a comment: We can look into how many of these apt commands (including from before this pull request) we really need. I have a branch where I removed almost every `apt-get update` and `apt-get install` from the `dockerfile` and it went fine.

The reason our setup can be so simple with regard to apt packages is that the (recently added) nodejs and yarn installation scripts download a whole lot of apt packages (like `build-essential`) that are generally useful later in the `Dockerfile`.

But the overall concept of using a slimmer base image sounds good IMO.

I think the criterion for merging this should be "does the app work" vs. errors during `docker-compose build` or `docker-compose up` or when testing the app.

Dockerfile
```diff
@@ -1,6 +1,11 @@
-FROM ruby:2.3.7
+FROM ruby:2.3.7-slim
 ENV PHANTOM_JS=2.1.1
 
+# Add basic binaries and clean up the apt cache
+RUN apt-get update \
+  && apt-get install -y bzip2 curl gnupg wget \
```
DeeDeeG: `apt-get install -y bzip2 curl gnupg wget`

These are super common, fundamental utilities that lots of stuff relies on. Probably need these by the end of the script. However, it'slikely these are provided during the `yarn` or `nodejs` install scripts. That makes this line redundant, and `apt-get` chews up time during `docker-compose`. So IMO we should test removing this if we want the speediest docker-compose times.

- `bzip2`: compression/decompression algorithm
- `curl`, `wget`: tiny utilities to download stuff off of the network
- `gnupg`: GNU Privacy Guard implements pgp cryptographic signing, helps verify stuff you downloaded hasn't been tampered with

DeeDeeG:
Hard drive space makes more of a difference if we were distributing our app somewhere in the form of pre-built Docker images, or using said hypothetical Docker image for production. But we are not doing that at the moment.

Dockerfile
DeeDeeG: This would have more impact after the yarn, phantomjs, and nodejs install scripts are done.

But I don't think we are targeting small final image size so much as fast builds. Not sure if rest of team agrees -- Are we going to start using the final built docker image for distributing the app, or in production? In which case having this could help final image size stay small.

Dockerfile
mi-wood: Yeah, I think it would be best to remove everything but the change in image and see if it works from there.

DeeDeeG: I realized just now we have installer scripts for `yarn` and `nodejs` thatwe download with `curl`. Likewise, we currently download the PhantomJS prebuilt package with `wget`. So:
- `curl` and `wget` need to stay.
  - OR we should have them both use `curl`, or both use `wget` and eliminate the download utility we don't use.

(Aside: to confirm this, I tried without `curl` and the build failed. I can try again without `wget`, but it's a bit tedious. The concept of this slim ruby base image is that it *doesn't* include most of these common utilities, whereas the regular ruby base image does include the super-common ones.)

I assume we do use `bzip2` and `gnupg`, but I'll try without. be back in a bit with the results.

DeeDeeG: So it looks like we need all four of these packages in order to install with the slim base image.

The rm `/var/lib/apt/lists/*` is totally optional, and just decreases the size of the built Docker image.

DeeDeeG:
I think we could copy some of the approach taken here: https://github.com/RefugeRestrooms/refugerestrooms/compare/develop...btyy77c:dockerAlpine

Noteworthy techniques in this branch:

- Install everything through a package manager (no downloading scripts or zip archives and installing from those)
  - use `yarn` and presumably `nodejs` from the linux distro's package repositories
    - e.g. `apt-get install nodejs` && `npm install yarn`
  - get PhantomJS-prebuilt with `yarn`
    - add `phantomjs-prebuilt` to the `Package.json`
- Drop lots of other stuff, since the package manager handles most of it for us.

DeeDeeG:
In terms of simply merging this PR, I would say we can remove the `rm /var/lib/apt/lists/*` command, or move it to after any `apt-get`commands. But everything else is pretty much good within the scope of this pull request.

For a fuller discussion of optimizing the Dockerfile, I opened #509.

darpr:
Thanks @DeeDeeG !

PR #508, opened 2018-10-08
Fix empty restrooms db during development

DeeDeeG:
Fixes db being empty during development.
Ensures restroom entries from `db/export.csv` are loaded.

# Context
- Copied the `bundle exec rake db:setup` command from the old [Vagrant configurations](https://github.com/RefugeRestrooms/refugerestrooms/blob/18ae19aeace01aa082cf548ae0818e490609f24e/setup/setup_vagrant.sh#L99).
- Makes it easier to test search and API pages
- Documentation/ reference:
  - This Stack Overflow answer: https://stackoverflow.com/questions/10301794/difference-between-rake-dbmigrate-dbreset-and-dbschemaload
  - The source code for these db functions: https://github.com/rails/rails/blob/v5.1.4/activerecord/lib/active_record/railties/databases.rake 

# Summary of Changes

- Remove `db:create` and `db:migrate` commands in `setup/entry`
- Replace with `bundle exec rake db:setup`

# Checklist

- [ ] Added Unit Tests
- [ ] CI Passes

# Screenshots

## Before

Restrooms search results showing no results and a mostly blank page

## After

Restrooms search results showing full page of restroom entries

### DIFF
--- a/setup/entry
+++ b/setup/entry
@@ -1,7 +1,6 @@
 #!/bin/bash
 set -e
 
-rake db:create
-rake db:migrate
+bundle exec rake db:setup
 
 exec "$@"
--- a/setup/entry
+++ b/setup/entry
@@ -1,6 +1,6 @@
 #!/bin/bash
 set -e
 
-bundle exec rake db:setup
+rails db:setup
 
 exec "$@"


DeeDeeG:
A more-minimal change might be to keep `db:create` and `db:migrate`, followed by `db:seed`.

I don't have time to test at this moment. But the `bundle exec rake db:setup` command does work in my testing.

DeeDeeG:
I'm hoping to merge this soon (in whichever configuration of commands we decide to go with) since it makes manual testing easier. It would particularly make the set of pull requests we have queued up easier to test.

btyy77c:
Thanks for this!

Some comments:

1. In Rails 5 they changed the command from `rake` to `rails`. You would need to use `rails db:setup` now.  I think `rake db:setup` will stop working with Rails 6:
https://github.com/rails/rails/blob/master/railties/lib/rails/tasks/initializers.rake
https://github.com/rails/rails/pull/33631

2. Just to confirm, this docker-compose file is not used for production, correct? `rails db:seed` will harm your production database.
If this is used for production, I would recommend instead running `docker-compose run web rails db:setup` in your terminal before you run `docker-compose up` to test things out locally... or creating a separate docker process for production. 

3. Your restroom model has a geocoding method:
https://github.com/RefugeRestrooms/refugerestrooms/blob/develop/app/models/restroom.rb
```
def perform_geocoding
      return true if Rails.env == "test"
      geocode
end
```
I think this is connecting to Google Api maps. When I run it, I get `Google Geocoding API error: over query limit.`  It's not a huge deal, just so you're aware things may not be loading the way you want.  I assume it's ok since you're skipping it for testing.

tkwidmer:
This LGTM @DeeDeeG, let's merge it asap since it seems like this is causing you and other people issues.

DeeDeeG:
Alright. Testing with the latest commit locally just to be sure this works, then will merge.

I think our test environment (and our tests themselves) don't need the `db/export.csv` entries, so that's why CI didn't catch this.

tkwidmer:
Awesome. Thanks @DeeDeeG

PR #510, opened 2018-10-09
_head.html.haml: Update Maps API version

DeeDeeG:
commit message:

> v3.31 is deprecated.
> 
> This also updates to the new weekly/quartery versioning scheme.
> See: https://developers.google.com/maps/documentation/javascript/versions

# Context
- The version of the Maps API we're using is deprecated, so it should be updated.
- We can stay up to date by using either `quarterly` or`weekly` as the version number.
  - See: https://developers.google.com/maps/documentation/javascript/versions
  - Versions such as `v3.34` are still usable, but it would be nice to never be out of date without having to worry about manual updates, in my opinion.
- There's release notes for all the versions of the Maps API here: https://developers.google.com/maps/documentation/javascript/releases

# Summary of Changes

- Changes our api version from `v3.31` to `quarterly`.

# Checklist

- [ ] Added Unit Tests
- [ ] CI Passes

# Screenshots

### DIFF
--- a/app/views/layouts/_head.html.haml
+++ b/app/views/layouts/_head.html.haml
@@ -6,7 +6,7 @@
 %meta{:content => "RefugeRestrooms.org", :name => "author"}/
 %meta{:content => "initial-scale=1, maximum-scale=1", :name => "viewport"}/
 
-%script{:src => "https://maps.googleapis.com/maps/api/js?v=3.31&libraries=places&key=AIzaSyBXcCrqzMlm-ZmtNQve7AuipNdE4vySUF0"}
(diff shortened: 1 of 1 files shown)


DeeDeeG:
I'm not sure which makes more sense for us:

- weekly updates
- quarterly updates

On the one hand, maybe we will miss out on some bug fix for up to three months. On the other hand, on the weekly update channel, we might get a bug one week that we have to hotfix by switching to another version.

I think quarterly is both safe (I don't think the Maps team would leave a channel buggy for three months) and not any slower than we already update. That's why I went with quarterly. If anyone has any other views and want to discuss, I am definitely open to that.

In any case, regardless of update schedule, it will be good to be on a version that's not deprecated.

DeeDeeG:
This is a pretty minor update, I think. Hoping to merge soon if there are no objections.

PR #513, opened 2018-10-12
Docker: Don't geocode the db seed entries

### LINKED ISSUE #512: Avoid geocoding when seeding the db during development
## Context

See this comment (bullet point number 3) by @btyy77c that brought this to my attention: https://github.com/RefugeRestrooms/refugerestrooms/pull/508#issuecomment-427796840

## Scope and Proposal

Not too hard, I hope.

- Add an environment variable (`SEEDING_DONT_GEOCODE`) or similar.
  - Probably set this in [`setup/entry`](https://github.com/RefugeRestrooms/refugerestrooms/blob/develop/setup/entry) before `rails db:setup`, then unset it right after that.
- Modify our controllers or models to check this environment variable and not geocode when environment variable (`SEEDING_DONT_GEOCODE`) is set.
  - Probably here: https://github.com/RefugeRestrooms/refugerestrooms/blob/develop/app/models/restroom.rb#L80
  - Or else here?: https://github.com/RefugeRestrooms/refugerestrooms/blob/develop/restrooms_controller.rb

## Impact

- Stops us from using [21 restroom submissions'](https://github.com/RefugeRestrooms/refugerestrooms/blob/develop/db/export.csv) worth of quota every time we bring up the Docker image.
- Eliminates error message:
  - `Google Geocoding API error: over query limit.`

DeeDeeG:
Commit message:

> They are already geocoded and do not need to be sent to the Maps API.
> 
> This saves quota usage, and prevents error messages when we are over quota limit.

# Context
- Fixes #512
- Avoid geocoding the `db/export.csv` restroom entries, since they're already geocoded, and because this saves quota usage.

# Summary of Changes

- Set a `SEEDING-DONT-GEOCODE` environment variable to "true" during the `setup/entry` Docker entry script
- Update `app/models/restroom.rb` to check for the `SEEDING-DONT-GEOCODE` environment variable and not geocode if that is set to "true"

# Checklist

- [ ] Added Unit Tests
- [ ] CI Passes

# Screenshots

## Before

before, showing database creation and lots of geocoding errors during server startup

## After

after, showing database creation without errors during server startup

### DIFF
--- a/app/models/restroom.rb
+++ b/app/models/restroom.rb
@@ -79,6 +79,7 @@ class Restroom < ApplicationRecord
 
     def perform_geocoding
       return true if Rails.env == "test"
+      return true if ENV["SEEDING_DONT_GEOCODE"]
       geocode
     end
 end
--- a/setup/entry
+++ b/setup/entry
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -e
 
-rails db:setup
+# Set up the database for development, and seed with placeholder data from `db/export.csv`
+SEEDING_DONT_GEOCODE="true" rails db:setup
 
 exec "$@"


DeeDeeG:
Alternate implementation with more info messages printed to console: https://github.com/DeeDeeG/refugerestrooms/tree/dont-geocode-when-seeding_with-console-messages

Let me know if that is preferred and I can update this pull request, or we can close and open a new PR, etc.

DeeDeeG:
This is failing in CI, because Code Climate wants me to add test coverage for this diff.

I can't think of anything to test programmatically here though. I'm not sure how a new spec would meaningfully test this functionality. If it's okay with folks, I'm leaning toward not trying to add a test to this. So we would just ignore that particular CodeClimate check.

DeeDeeG:
Thanks for taking a look at this!

I made a branch with your suggested change.

I also:

- simplified the comment lines to just one in `setup/entry`
- Instead of checking if `SEEDING-DONT-GEOCODE` equals the string "true", just check if it is set at all
  - (I think string comparisons are slower than boolean checks, so I guess if we ever batch-geocode, all the string comparisons of "true" against "true" could slow down the server. But regardless, better fast than slow if it acts the same.)

https://github.com/RefugeRestrooms/refugerestrooms/compare/develop...DeeDeeG:dont-geocode-when-seeding-2

If this looks good, I will push that version to this branch so we can merge it.

DeeDeeG:
I am going to update this as mentioned above.

(The original commit from this Pull Request is backed up here: https://github.com/RefugeRestrooms/refugerestrooms/compare/develop...DeeDeeG:dont-geocode-when-seeding-db-during-development-old)

DeeDeeG:
Going to merge. Thanks for the comments!

PR #515, opened 2018-10-13
Optimize Travis CI build speed

DeeDeeG:
Commit message: 

> Moves `docker-compose build`
> from the `before_install:` step to the `install:` step
> in Travis CI.
> 
> This overrides the default `install:` step for Ruby apps,
> which is `bundle install --jobs=3 --retry=3`.
> 
> In doing so, we avoid running `bundle install` twice:
> - Once inside the Docker container (good)
> - Again outside of the container (unnecessary)
> 
> (The `bundle install` outside of the Docker container
> would have no bearing on our web app, tests, etc...
> but it would make Travis CI take longer to build the branch.)
> 
> See:
> https://docs.travis-ci.com/user/customizing-the-build/#the-build-lifecycle
> https://docs.travis-ci.com/user/customizing-the-build/#customizing-the-installation-step
> https://docs.travis-ci.com/user/customizing-the-build/#breaking-the-build
> 
> https://docs.travis-ci.com/user/languages/ruby#dependency-management

# Context
- Shortens Travis CI runs
  - Avoids running `bundle install` outside of the Docker container (where it has has no bearing on our app)

# Summary of Changes
- Overrides default `install` step:
  - Moves `docker-compose build` to the `install` step instead of `before_install`.

# Checklist

- [ ] Added Unit Tests
- [ ] CI Passes

# Screenshots

## Before
before-travis

## After
after-travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ env:
     - CC_TEST_REPORTER_ID=c18df080592f9c99ca8080a6d5e052aa5fd3964044a0fe0b71e48f8e18998dc2
 language: ruby
 services: docker
-before_install:
+install:
   - docker-compose build
 before_script:
   - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter


DeeDeeG:
This works as expected. It's nice to have faster Travis builds, so merging.

PR #518, opened 2018-10-14
Store reverse geo-coded address & force unique addresses

### LINKED ISSUE #399: Prevent duplicate locations when adding bathrooms
Currently, duplicate bathrooms can be submitted due to very small differences in addresses.
For example: In Bellingham, Wa the restroom in The Black Drop Coffeehouse appears twice because one address is listed as
300 West Champion
and the other restroom as
300 W Champion
There is only one restroom at this location, but it is listed as two separate restrooms, with two separate ratings because of the difference of W vs. West.

Further complications are made by BELLINGHAM vs. Bellingham and WA vs. Washington

These should be a simple fix by reducing all city names to lowercase for comparison and string parsing for W, E, N, S, etc. to their corresponding West, East, North, South, etc.

tcmal:
# Context

- Fixes #399

Generally, this makes the data more consistent, however this relies more on the geocode lookup being correct for what the user puts in. It might be a good idea to force the user to preview (via google maps) before they can submit a restroom, or to use a proper address lookup. Thoughts?

# Summary of Changes

Restrooms will now populate the geocode, then set the address fields to the address from a reverse lookup before validation.

The address field must be unique. This is only enforced in the activerecord model & not the database schema, meaning it won't complain about existing records / manually added records.

Added a task that iterates over each restroom and re-populates the address from a reverse geocode lookup (`db:reverse_geocode` / `db:reverse_geocode[dry_run]`)

### DIFF
--- a/app/models/restroom.rb
+++ b/app/models/restroom.rb
@@ -18,9 +18,10 @@ class Restroom < ApplicationRecord
   ignoring: :accents
 
   validates :name, :street, :city, :state, presence: true
+  validates :street, uniqueness: {scope: [:city, :state], message: "is already registered"}
 
   geocoded_by :full_address
-  after_validation :perform_geocoding, :reverse_geocode
+  before_validation :perform_geocoding, :reverse_geocode
 
   reverse_geocoded_by :latitude, :longitude do |obj, results|
     if geo = results.first
--- a/app/models/restroom.rb
+++ b/app/models/restroom.rb
@@ -21,7 +21,8 @@ class Restroom < ApplicationRecord
   validates :street, uniqueness: {scope: [:city, :state], message: "is already registered"}
 
   geocoded_by :full_address
-  before_validation :perform_geocoding, :reverse_geocode
+  before_validation :perform_geocoding, if: ->(obj){ obj.full_address.present? }
+  before_validation :reverse_geocode, if: ->(obj){ obj.full_address.present? }
 
   reverse_geocoded_by :latitude, :longitude do |obj, results|
     if geo = results.first
--- a/app/models/restroom.rb
+++ b/app/models/restroom.rb
@@ -20,11 +20,10 @@ class Restroom < ApplicationRecord
   validates :name, :street, :city, :state, presence: true
 
   geocoded_by :full_address
-  after_validation :perform_geocoding
+  after_validation :perform_geocoding, :reverse_geocode
 
   reverse_geocoded_by :latitude, :longitude do |obj, results|
     if geo = results.first
-      obj.name    = geo.address
       obj.street  = geo.address.split(',').first
       obj.city    = geo.city
       obj.state   = geo.state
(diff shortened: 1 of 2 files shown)


DeeDeeG:
Hi @tcmal,

Thanks for this contribution. I am going to cc @tkwidmer or @mi-wood to review this, since I don't usually deal with the DB. I will say, this seems like a neat premise!

And... Happy Hacktoberfest! :jack_o_lantern: :computer: :sparkles: :fallen_leaf:

lib/tasks/reverse_geocode.rake
```diff
@@ -0,0 +1,26 @@
...
+
+    if args.dry_run
+      puts "Dry running reverse_geocode"
+    else
+      puts "Running reverse_geocode"
+    end
+
+    puts ""
+
+    Restroom.all.each do |restroom|
```
cllns: Do we want to do this for _all_ Restrooms, or can we just do ones that have some columns missing?

lib/tasks/reverse_geocode.rake
```diff
@@ -0,0 +1,26 @@
...
+  task :reverse_geocode, [:dry_run] => [:environment] do |t, args|
+    args.with_defaults(dry_run: false)
+
+    if args.dry_run
+      puts "Dry running reverse_geocode"
+    else
+      puts "Running reverse_geocode"
+    end
+
+    puts ""
```
cllns:
```suggestion
    puts
```

app/models/restroom.rb
```diff
@@ -18,13 +18,14 @@ class Restroom < ApplicationRecord
   ignoring: :accents
 
   validates :name, :street, :city, :state, presence: true
+  validates :street, uniqueness: {scope: [:city, :state], message: "is already registered"}
 
   geocoded_by :full_address
-  after_validation :perform_geocoding
+  before_validation :perform_geocoding, if: ->(obj){ obj.full_address.present? }
+  before_validation :reverse_geocode, if: ->(obj){ obj.full_address.present? }
```
cllns: I think you can DRY this and make it a little simpler by using a method here.
https://guides.rubyonrails.org/active_record_validations.html#using-a-symbol-with-if-and-unless

```ruby
before_validation :perform_geocoding, if: :require_geocoding?
```
```ruby
def require_geocoding?
  full_address.present?
end
```

Also, `full_address` should be reimplemented, since right now it always returns a string (so this condition is always `true`. 

Instead of:
```ruby
def full_address
  "#{street}, #{city}, #{state}, #{country}"
end
```

```ruby
def full_address
  [street, city, state, country].compact.join(",")
end
```
(I think this is better than showing empty values, but "123 Main Street, USA" doesn't really make sense either.)

lib/tasks/reverse_geocode.rake
tcmal: Depends if you're particularly fussy about the display, for example casing and `Street` vs `St.`

DeeDeeG:
By the way, when I looked at this, it made the `SEEDING_DONT_GEOCODE` thing I did in #513 not work.

I'm not fluent with Ruby enough to get why, although I can tell a lot is changed around in the same file that #513 touched, so it makes sense.

(Not sure how important it is, probably not quite enough to block this new feature, but if I can help make it work again, would be happy to.)

PR #520, opened 2018-10-26
Prevent upvoting or downvoting multiple times on the same restroom

### LINKED ISSUE #357: Prevent the same person from rating more than once
There's no good way to do this completely, but I think it's good to set a cookie so the average user can't upvote or downvote multiple times.

translatorzepp:
# Context
- Fixes #357 
- Basic prevention of double-voting for a restroom.

# Summary of Changes

- After up- or downvoting a restroom, store that restroom ID in `session` and prevent tallying further votes for that restroom
- Add a flash message in English and Spanish

## Concerns

- I left off Javascript changes to disable the buttons, but I'd be happy to add it in if folks think it would be nice!
- I think we're safe to put database IDs into session since it's encrypted, but if anyone has concerns about leaking information, let me know.

# Checklist

- [x] Added Unit Tests
- [ ] CI Passes

# Screenshots

## After
already-voted

### DIFF
--- a/app/controllers/restrooms_controller.rb
+++ b/app/controllers/restrooms_controller.rb
@@ -39,9 +39,9 @@ class RestroomsController < ApplicationController
 
   def update
     if params[:restroom][:downvote]
-      Restroom.increment_counter(:downvote, @restroom.id)
+      _vote_for_restroom(:downvote)
     elsif params[:restroom][:upvote]
-      Restroom.increment_counter(:upvote, @restroom.id)
+      _vote_for_restroom(:upvote)
     elsif @restroom.update(permitted_params)
       flash[:notice] = I18n.t('restroom.flash.updated')
     else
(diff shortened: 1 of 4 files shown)


DeeDeeG:
Thank you for this!

Hopefully @mi-wood or @tkwidmer can take a look at this, since they have dealt with this part of the code more than me. But on first look, this appears pretty good!

I'm not familiar with the Rails session concept, so I'll perhaps take a look into that before merging.

In any case... Happy Hacktoberfest! :jack_o_lantern: :computer: :sparkles: :fallen_leaf:

DeeDeeG:
Here are some resources about the Rails session feature...

Tutorials, explainers and how-to's:
https://www.justinweiss.com/articles/how-rails-sessions-work/
https://www.theodinproject.com/courses/ruby-on-rails/lessons/sessions-cookies-and-authentication
https://www.tutorialspoint.com/ruby-on-rails/rails-session-cookies.htm

Official docs:
https://guides.rubyonrails.org/action_controller_overview.html#session
https://guides.rubyonrails.org/security.html#sessions

PR #525, opened 2018-11-07
Ruby 2.5

### LINKED ISSUE #523: Update Ruby (the Ruby 2.3 series hits end-of-support in April 2019)
Hi folks,

According to Ruby-lang.org, [Ruby 2.3 is in its "security maintenance phase."](https://www.ruby-lang.org/en/news/2018/10/17/ruby-2-3-8-released/)

> Ruby 2.3 is now under the state of the security maintenance phase, until the end of the March of 2019. After the date, maintenance of Ruby 2.3 will be ended. We recommend you start planning migration to newer versions of Ruby, such as 2.5 or 2.4.

## Impact

The best security, and the most current feature updates are available on more-recent Ruby. It would theoretically be a plus to move to newer series of Ruby.

## Proposed solution

We can update to Ruby 2.4, or Ruby 2.5.

For details on these: https://www.ruby-lang.org/en/news/

## How to actually do this

Here is a general-purpose guide to updating Ruby in a Rails application: https://everydayrails.com/2017/12/18/ruby-upgrade-guide-for-rails.html

However, for us it is a bit simpler, since we do development in containers/VMs...

- I personally would just do a `git grep "2\.3\.7"` and replace every instance of our current ruby version with the new version, `[maj].[min].[patch]`.

- Then we can run the tests and/or generally troubleshoot. I suspect an upgrade to 2.4 would go rather smoothly at this point, and 2.5 might be bumpier, but the only way to know is to try it out.

DeeDeeG:
# Context
- Fixes #523
- Update Ruby (and the Webmock gem).

# Summary of Changes

- Update all references to a specific Ruby version from `2.3.7` --> `2.5.3`
- Update `README.md` from saying we use `2.3.0` to saying we use `2.5.3`
- Update `webmock` gem (the new version of webmock is compatible with Ruby 2.5; old version has errors when used with Ruby 2.5 even though it worked fine with 2.3.7)

# Checklist

- [ ] Added Unit Tests
- [x] CI Passes
- [x] Deploys to Heroku on test Correctly (Maintainers will handle)

# Screenshots


DeeDeeG:
I did a test deploy of this.

It seems to work fine.

...

Bit of a red herring, but worth explaining: In general, when I have deployed to (my personal account at) Heroku recently, the "Add a new restroom" page breaks. This happens with Ruby 2.3 (See: https://refuge-cedar-14.herokuapp.com/) or Ruby 2.5 (See: https://refuge-ruby-25.herokuapp.com/).
And I used to get so many errors when I tried to deploy to my personal Heroku account that the whole deploy was just error messages. Meanwhile, our normal deploy process for production has always gone better than that.

So I do think it's an issue with how _I'm set up with/using Heroku_, rather than a problem with our code or our actual deploy process for production.

DeeDeeG:
The deploy failures here were when I had the "ruby" buildpack before the "nodejs" buildpack.

If the nodejs buildpack isn't first, it won't run at all... So it has to be "nodejs" then "ruby" in the buildpack configs. Which is what I believe we already do (the right way) in production.

DeeDeeG:
I've had no problems with this so far, so going to merge in a few hours if there are no objections.

DeeDeeG:
Should this wait until after ActiveAdmin and Kaminari are bumped? Other wise I'm doing this after the kilometer/miles translations are added.

mi-wood:
Go for it. I'm not sure I'm going to have time to fix the kaminari issue for a bit.

PR #529, opened 2018-11-18
399 duplicates

mi-wood:
This is just a slight refactor of the PR I merged to a local branch. Moving back to using `(obj)` syntax, because it seems like the methods aren't available in the `if` block.

### DIFF
--- a/app/models/restroom.rb
+++ b/app/models/restroom.rb
@@ -21,8 +21,8 @@ class Restroom < ApplicationRecord
   validates :street, uniqueness: {scope: [:city, :state], message: "is already registered"}
 
   geocoded_by :full_address
-  before_validation :perform_geocoding, if: require_geocoding?
-  before_validation :reverse_geocode, if: require_geocoding?
+  before_validation :perform_geocoding, if: ->(obj){ require_geocoding? }
+  before_validation :reverse_geocode, if: ->(obj){ require_geocoding? }
 
   reverse_geocoded_by :latitude, :longitude do |obj, results|
     if geo = results.first
@@ -46,6 +46,10 @@ class Restroom < ApplicationRecord
 
   scope :created_since, ->(date) { where("created_at >= ?", date) }
   scope :updated_since, ->(date) { where("updated_at >= ?", date) }
+
+  def require_geocoding?
+    full_address.present?
+  end
 
   def full_address
     [street, city, state, country].compact.join(",")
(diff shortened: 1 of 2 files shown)


mi-wood:
Looks like we need to stub the geocode require that the if block creates. I don't have time to do this atm, but can get to it later. Also, not really sure if there's a better way to fix the ugly `->(obj)` block syntax.

app/models/restroom.rb
```diff
@@ -17,14 +17,17 @@ class Restroom < ApplicationRecord
   using: {tsearch: {dictionary: "english"}},
   ignoring: :accents
 
-  validates :name, :street, :city, :state, presence: true
+  if Rails.env != "test"
```
mi-wood: This does the trick for testing, but I don't like it. I think I'd rather fix the tests to truncate the db and simulate production better. We also need a test for the validation itself.

db/seeds.rb
```diff
@@ -21,7 +21,8 @@
       :comment => row[12],
       :latitude => row[8],
       :longitude => row[9],
-      :country => row[6]
+      :country => row[6],
       )
+    restroom.save(validate: false)
```
mi-wood: Ignoring validation during seeding seems reasonable, since those don't really change ever.

app/models/restroom.rb
DeeDeeG: Would we be able to "simulate production better" by having real database data from production? I would be able to update `db/export.csv` with actual, recent data as it exists in production, if that would help.

mi-wood: @DeeDeeG I think that would make sense. I'd like to also work on killing this conditional though.

mi-wood:
This should also have a test for it. @tcmal do you think you could add one?

app/models/restroom.rb
mi-wood: Oh, also FYI. Changing the csv won't fix this issue. There's some sort of hook before the save that tries to geolocate these. I'm not sure what forces it to happen, but it is different than the uniqueness check.

tcmal:
Will do, in a fork or just put the patch on a gist?

mi-wood:
Fork and a PR into this branch would be great! Sorry if I made that a pain. It’s tough to work on multiple forks.

DeeDeeG:
@mi-wood we can go with whatever your preference is, but if you have no preference, and in general, I'd be comfortable accepting the changes either way (patch or Pull Request). If we get it as a patch, I or @mi-wood may as well commit that to this branch/this Pull Request.

cc @tcmal

DeeDeeG:
For the sake of the email notification chain, reposting this comment by @mi-wood:

>Fork and a PR into this branch would be great! Sorry if I made that a pain. It’s tough to work on multiple forks.

DeeDeeG:
The merge conflict here is that two branches each added a different test to the repo's RSpec test suite. Best-case, both tests work as intended in this branch and we just keep them both.

But if these conflict with each-other somehow, or one of the tests breaks in this branch, then I guess we have to adjust or delete one of the tests, and/or worst-case, one of the features being tested...

PR #531, opened 2018-11-18
Release

mi-wood:
This is a release for:
- Hindi translation
- User edit process

PR #533, opened 2018-11-20
Revert "Release"

mi-wood:
Reverts RefugeRestrooms/refugerestrooms#531


DeeDeeG:
It would be nice to merge the Hindi translation in again, if we can do that.

(And between the two things, I'd have to guess the user edit proposals is more likely to cause memory issues than a new translation would, not that everything intuitive turns out to be true with computers....)

mi-wood:
Closing, because I patched it instead.

PR #565, opened 2019-03-21
Add a "bug report" issue template

DeeDeeG:
Gives a more reasonable template for submitting bugs.

(The "Generic Proposal" template doesn't really fit for bug report issues.)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug--something-went-wrong-or-isn-t-working-right-.md
+++ b/.github/ISSUE_TEMPLATE/bug--something-went-wrong-or-isn-t-working-right-.md
@@ -1,0 +1,26 @@
+---
+name: Bug (something went wrong or isn't working right)
+about: For bug reports, or describing problems with Refuge Restrooms
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+### What isn't working
+
+<!-- Please describe here what isn't working right. -->
+
+<!-- For example: 
+
+### What usually happens
+
+<!-- What would you usually expect to happen instead? -->
+
+### Steps to recreate the problem
+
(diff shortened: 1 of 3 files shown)


DeeDeeG:
I was thinking whether to make a bug report template, or just a blank template. I think people have been pretty good at describing issues in the past, and it's probably nicer to do follow-up questions in the issue comments than to make contributors jump through too many hoops.

Anyway, I drafted this to get some comments/feedback.

I'd still be okay with just adding a blank issue template, which I think would help folks feel they don't need to use the generic proposal template when it doesn't really fit.

DeeDeeG:
HMM there's three of these... Probably because I changed the title of the issue template a few times while drafting it. I can fix this by deleting two of them and force-pushing to the branch...

tkwidmer:
Whats the status of this?

DeeDeeG:
Hi @tkwidmer,

I wasn't able to decide which of these I like. They saved at different times when I was drafting them, at the moment I updated the name.

Simpler one: https://gist.github.com/DeeDeeG/0d18e4b845623fe5bac43f19c956c77b#file-simple-template-md
More complex one: https://gist.github.com/DeeDeeG/0d18e4b845623fe5bac43f19c956c77b#file-complex-template-md

I might also lean toward the default GitHub one: https://gist.github.com/DeeDeeG/0d18e4b845623fe5bac43f19c956c77b#file-github-default-template-md

... Or just one called "blank" so folks can write what they want.

Kind of hard to decide. But it's a bit of one-person [bikeshedding](https://en.wikipedia.org/wiki/Bikeshedding) I guess.

(If need be, this can just be closed.)

DeeDeeG:
Closing this for now.

Issue #172, opened 2014-10-28
Implement Google's No Captcha


tkwidmer:
@backspace implemented Akismet to cut down on the amount of spam we were receiving. But it is not open source / free software. So we want to switch to an open source solution in the future.

mi-wood:
Just adding a note, that this would be incredibly helpful for the contact form. We get a lot of spam

hendrixfan:
How about using [Invisible Captcha](https://github.com/markets/invisible_captcha)? It is Open Source and uses the honeypot Princible.

DeeDeeG:
@hendrixfan that could be a good idea! Probably better than having to get ReCaptcha over the network! That's one of the least-performant aspects I noticed of the site, mostly in terms of latency/full-page load-time. I think the ReCaptcha code is also pretty large over the network next to other things users don't have to re-download once they've cached a part of the site once (after the first page visit).

We recently did finally add ReCaptcha, but it could be good to replace it with this open-source solution. cc @mi-wood curious if this open-source alternative looks good enough, or if there'd still be too much spam? As I know you've been the one to deal with the spam the most.

PR #566, opened 2019-03-23
Implement Google's reCAPTCHA

stardust66:
# Context
- Implements #172.
- Add reCAPTCHA (checkbox style) to contacts and restroom submission page.

# Summary of Changes

- Add reCAPTCHA to contacts and restroom submission forms.
- Verify reCAPTCHA in respective controllers.
- Enable form validation by default.
- New localized string in `config/locales/en/simple_form.en.yml` for reCAPTCHA failure message. `en.helpers.reCAPTCHA.failed`

### Note
I got a site key and a secret key for testing, and I put them in `.env`, which gets loaded by the `dotenv` gem. An admin should register the site with Google (https://www.google.com/recaptcha/admin/create) and put the actual site key and secret key in the Heroku env variable settings.

# Screenshots
Screen Shot 2019-03-23 at 1 50 05 PM

image

image


DeeDeeG:
I may not be an authority on this, but I'd say the code looks solid. +1 on technical implementation.

(I can read Ruby better than I can write it at this point.)

PR #567, opened 2019-03-28
db/seeds.rb: Give restroom entries an edit_id

### LINKED ISSUE #541: Data from db/export.csv should get "edit_id" values matching their "id" values (currently all set to zero)
## Scope / difficulty of fixing this

Not too hard I hope?

## Impact

Only affects local testing (in Docker, Vagrant, etc.) with the `db/export.csv` data. Fixing this will make the data from `db/export.csv` behave properly alongside the new user edit proposals system.

_This has no bearing on our production instance, because the data there all have proper`edit_id` values._

## Problem

Data in `db/export.csv` have no actual `edit_id` values to begin with. 

Rather than being migrated (or otherwise automatically processed) to have an `edit_id` matching their `id`, they all end up having `edit_id: 0`. 

Since we now group API results by `edit_id` and only display one result per `edit_id`, the app only ever displays at most one result regardless of the specifics of the API call. (This will tend to be the Dolores Park Cafe entry, since it has the highest `id` value).

(From the documentary about software development, _Highlander_: **"There can be only one!"**)

## How to reproduce the problem

- Run the app locally (for example in Docker or Vagrant)
- Search for restrooms near "San Francisco" anywhere in the app
  - or visit the API documentation page and make any API call
- Observe there are either no relevant results (not related to this bug) or else only one result (because of this bug).

## Proposed solution to this bug

Process the `db/export.csv` data in some way (a migration? somewhere else in our Ruby code?) so they have `edit_id` values matching their `id` values.

DeeDeeG:
# Context
- Fixes #541
- This issue was that all the data in development environment were grouped together (as if they were serial edits of the same original restroom), and so only one restroom entry was being displayed.

# Summary of Changes

In `db/seeds.rb`:
- For each entry being generated from the .CSV file, copy the generated `id` value to also be the `edit-id` value. 
  - This ends up giving a matching `id` and `edit_id` to every entry.

[Edited the above: This no-longer count up the `edit_id` number from 1 by iterating the loop; It now reads the `id` value of the restroom entry, and copies that as the (matching) `edit_id` value.]

# Checklist

- [ ] Added Unit Tests
- [x] CI Passes

# Screenshots

## Before

Refuge Restrooms API documentation page, showing only one restroom entry with edit_id: 0

## After

API documentation page, showing multiple restroom entries, where each entry has matching edit_id and id values

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -9,8 +9,8 @@
 require 'csv'
 
 Restroom.transaction do
-  CSV.foreach('db/export.csv').with_index(1) do |row, i|
-    Restroom.create(
+  CSV.foreach('db/export.csv') do |row|
+    restroom = Restroom.create(
       :name => row[1],
       :street => row[3],
       :city => row[4],
@@ -21,8 +21,8 @@ Restroom.transaction do
       :comment => row[12],
       :latitude => row[8],
       :longitude => row[9],
-      :country => row[6],
-      :edit_id => i
+      :country => row[6]
       )
+    restroom.update(edit_id: restroom.id)
   end
 end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -9,7 +9,7 @@
 require 'csv'
 
 Restroom.transaction do
-  CSV.foreach('db/export.csv') do |row|
+  CSV.foreach('db/export.csv').with_index(1) do |row, i|
     Restroom.create(
       :name => row[1],
       :street => row[3],
@@ -21,7 +21,8 @@ Restroom.transaction do
       :comment => row[12],
       :latitude => row[8],
       :longitude => row[9],
-      :country => row[6]
+      :country => row[6],
+      :edit_id => i
       )
   end
 end


mi-wood:
I think it might be better to do something similar to how the application creates restrooms. So:

```
restroom = Restroom.create(...)
restroom.update(edit_id: restroom.id)
```

That way it is tied to the actual `id` and not the index of the for loop.

DeeDeeG:
Thanks for the code suggestion, looks good.

I got kind of confused trying to add that in, because using the wrong thing (`restroom` vs `Restroom`) did completely different things. For clarity, would you mind if I did this:

```ruby
current_restroom = Restroom.create(...)
current_restroom.update(edit_id: current_restroom.id)
```

Or something similar?

mi-wood:
So, `Restroom` is a class name. 

`Restroom.create` is a class method. This is a method is not associated with a single restroom, hence it is used for `create` (the restroom doesn't exist yet). It returns an instance of `Restroom`. 

`restroom.update` is an instance method. It acts upon a restroom. In this case, `update` is acting upon the existing restroom to get its `id` and update the `edit_id`.

`class_name = ClassName.new()` is a pretty common convention. That being said, I'm not too particular about what it's called.

DeeDeeG:
I went with just lowercase `restroom` for the instance name, because, as you say, it is a common pattern in Ruby, and people editing this file are likely to be comfortable with common Ruby patterns.

Going with the flow is fine here IMO.

(This can be squash-and-merged for neatness' sake in `develop`/`master`.)

DeeDeeG:
I'm going to merge this soon if that's okay/no objections. (Definitely want to squash&merge though.)

PR #570, opened 2019-03-30
Explain how to run individual tests and access psql

purplejacket:
Also, clarify that there are two containers, web and db,
rather than just one; these can be viewed using
docker ps

# Context
- Fixes #[issueNumber]
- short_summary_here

# Summary of Changes

- change_1_here
- change_2_here

# Checklist

- [ ] Added Unit Tests
- [ ] CI Passes

# Screenshots


lisavogtsf:
I like these suggestions! I want to write even more about the `docker-compose` interactions and troubleshooting, but who knows if I'll get to it. For what it's worth, I approve!

Issue #428, opened 2018-01-19
Make a language switcher for the site


DeeDeeG:
Hi all,

In discussions about how best to show users the various languages we will have available, now that we are getting translations merged in, it was mentioned that a language switcher has advantages over purely automatic system. [[1](https://github.com/Intimaria/refugerestrooms/pull/1#issuecomment-357035097)] 

Basically, a user may be automatically detected as needing one locale, but they may prefer or be better suited by using a different locale. The language switcher allows them to pick the best language (of the ones we have) that suits their needs.

#### Language Selection Logic

I think our current logic for determining the locale to use should be expanded upon.

It currently just does this: "Check the user's accept-language HTTP headers", then "If we have one of those languages, or even their exact lang-REG locale, serve it." -- "Otherwise, serve the `en` locale."

I propose we add a step, right at the beginning:

"Read the 'lang' parameter of the URL", then "If we have that locale, serve it", otherwise fall back to the previously described system. 

(The other major option seems to be using different lang subdomains, e.g. `es.refugerestrooms.org`, `zh.refugerestrooms.org`.) [[2](http://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests)]

#### The switcher itself -- Setting the "lang=" parameter of the URL

This means we have to add the switcher, in a way that can set the language parameter of the URL. We also need a way to persist or carry-over that language parameter of the URL when navigating to other sub-pages of the site.

Since the language switcher interface should always be visible, I propose we should add it in the footer. (We can make the switcher its own `_partial.html.haml` and render it in that way.)

I expect all the code for this would be either pure Ruby, or a mix of Ruby and JavaScript. JavaScript code can be embedded in Ruby code, so if it's mixed, I think the code would all be in a `.rb` file.

We can then evaluate this from the `.html.haml` file with `= _lang_switcher`, or something along those lines.

#### User Interface

(Proposing a 'language' drop-down menu near the bottom (or top?) of the page. We can make it look nice with a flag image next to each locale, and format it with CSS, when the time comes.

But anything that works and looks nice or has a good user-experience is welcome as a suggestion.)

Open to discussion, and particularly if this sounds like a good plan, we may get going on an implementation.

DeeDeeG:
There is [this](https://rubygems.org/gems/language_switcher) [gem](https://github.com/teonimesic/language_switcher) to implement a language switcher. I don't know if it fits our needs, but it could be worth giving it a try.

[...]

PR #585, opened 2019-04-27
Add "locale=" URL parameter to the webapp (prepares for adding a locale switcher)

DeeDeeG:
# Context
- Partly addresses #428
- Adds a URL parameter `?locale=[your_locale_here]`
  - Prepares app to use the parameter, try to fix hard-coded links that don't have the parameter.

# Summary of Changes

- Add a `?locale=[locale]` default URL parameter to all dynamically-generated links, e.g. links made with the Rails `link_helper` 
- Adjust locale selection to, during page load, look for this `?locale=` parameter, and if present, use that parameter's string as the locale. 
  - (Thus far, locale has only been set from the user's browser locale preference list. There has been no way to specify what language Refuge appears in beside editing the browser locale preferences.)
    - Code in this PR still falls back to the browser locale list if no `?locale=` parameter is present in the URL.
- Initial attempt to track down hard-coded links or redirects, and replace with dynamically generated ones such as with Rails `link_helper` code.
  - So far only adjusted the main homepage link in the far corner of the topnav area.
  - The search box still redirects in a way that does not include the `?locale=` parameter at this time.
    - This should be fixed, or else the search box will break the `?locale=` chain, and make a language switcher seem "broken" and be much less useful.

# Checklist

- [ ] Added Unit Tests
- [ ] CI Passes

# Screenshots

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,7 +10,7 @@ class ApplicationController < ActionController::Base
   end
 
    def set_locale
-     I18n.locale = http_accept_language.language_region_compatible_from(I18n.available_locales)
+     I18n.locale = params[:locale] || http_accept_language.language_region_compatible_from(I18n.available_locales)
    end
 
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -13,4 +13,8 @@ class ApplicationController < ActionController::Base
      I18n.locale = params[:locale] || http_accept_language.language_region_compatible_from(I18n.available_locales)
    end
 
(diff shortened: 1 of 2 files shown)


DeeDeeG:
See this documentation for details: https://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests

There may be a nicer way to do this, i.e. with sub-directories for each locale: `https://refugerestrooms.org/en/`, `https://refugerestrooms.org/es`, `https://refugerestrooms.org/pl` etc.

DeeDeeG:
Closed by #620. :tada:

PR #586, opened 2019-04-27
Dockerfile: Update and streamline steps

### LINKED ISSUE #509: Simplify and streamline the Dockerfile
## Scope / difficulty

Just need to update the [`Dockerfile`](https://github.com/RefugeRestrooms/refugerestrooms/blob/develop/Dockerfile). Some discussion beforehand for consensus would be great.

## Impact

Would make `docker-compose build` run faster, and maybe make things easier to maintain; Should help the Docker config be more readable and easier to work with. Makes dependencies more obvious and more apparent what things are for what purpose.

## Rationale

I realized while working on a [branch](https://github.com/DeeDeeG/refugerestrooms/commits/vagrant-bionic-with-rvm-this-works-Oct-2018) that makes Vagrant usable (again, I'm not sure why I can't let go of Vagrant 🤷) that we can [simplify](https://github.com/DeeDeeG/refugerestrooms/commit/cb258ad3dbd9a71f206559340cf0938eec583acb) the Docker script. A lot of `apt-get update` and `apt-get install` in the script can be removed, now that we have the nodejs, phantomjs, and yarn install scripts that install most or all of those packages regardless.

I also noticed @btyy77c's branch over at https://github.com/btyy77c/refugerestrooms/tree/dockerAlpine, that shows how much neater the script can get. I think we might want to ask for some of those things as a Pull Request, or get the okay to re-implement some of those changes. (Would want to give credit in the commit messages.)

## Proposal

I think we should try some simplified `Dockerfile`s and discuss which is the best direction to go.

OR just make a pull request using some progress or improvements from the aforementioned branches:
- https://github.com/DeeDeeG/refugerestrooms/tree/cleanup-Dockerfile
- https://github.com/btyy77c/refugerestrooms/tree/dockerAlpine
And discuss there.

## How to actually do this:

- Consider improvements that can be made to the `Dockerfile`. See branches that are linked above for ideas.
- Probably merge #504.
  - It's a little late, but we could consider https://github.com/btyy77c/refugerestrooms/tree/dockerAlpine as an alternative to #504, but I think that would be unkind this late in review for that PR. Another downside is, we would be working with Alpine, which is a tad less familiar than Debian (which in turn is almost identical to working with Ubuntu). But we don't need to mess with the `Dockerfile` very often, so I don't think that's a big issue.
- Latstly, implement improvements by submitting the updated `Dockerfile` as a pull request.

OR

- Same as above, but start with the pull request and discuss and potentially make changes from there.

DeeDeeG:
Does effectively the same things as before,
but now in a simpler/faster way.

Some of the changes take inspiration from
`@btyy77c`'s dockerAlpine branch:
https://github.com/btyy77c/refugerestrooms/blob/dockerAlpine/Dockerfile

The PhantomJS install is based on (mostly copy-pasted from)
`@nkovacs`' phantomjs image from Docker Hub:
https://github.com/nkovacs/selenium-standalone-phantomjs/blob/c5f6bba218472270/Dockerfile#L19-L22

# Context
- Fixes #509
- Optimize the Dockerfile for install time, size on disk, network usage... and readability

# Summary of Changes

- Looked at each step and replaced with simplest possible re-implementation, without causing headaches for usability or maintainability/stability.
- Install way fewer apt packages
- Avoid using installer scripts where we can just download a zip or tar archive and extract. (e.g. Node, PhantomJS)
- Install Yarn with `npm install -g yarn` to save time, network usage, disk usage...
- Use `yarn install --frozen-lockfile` rather than `yarn --pure-lockfile`; If our Node dependencies are broken, I think we want to know about it rather than have it silently install similar (but not exact) dependencies?? Also using the word `install` is more explicit/verbose and therefore more readable, in my personal opinion.

# Checklist

- [ ] Added Unit Tests
- [x] CI Passes
- [x] Added Documentation (Service and Code when required)

# Screenshots

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,38 +1,43 @@
 FROM ruby:2.5.3-slim
-ENV PHANTOM_JS=2.1.1
 
 # Add basic binaries
 RUN apt-get update \
-  && apt-get install -y bzip2 curl gnupg wget
+  && apt-get install -y curl g++ gcc libfontconfig libpq-dev make patch xz-utils \
+  # Clean up the apt cache
+  && rm -rf /var/lib/apt/lists/*
 
-# Add the apt repository for yarn
-RUN curl -sS http://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
-  echo "deb http://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+# Download, extract and install PhantomJS from archive hosted at bitbucket
+RUN curl -L https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.1.1-linux-x86_64.tar.bz2 -O \
+  # Extract and clean up the PhantomJS archive
+  && tar xf phantomjs-2.1.1-linux-x86_64.tar.bz2 && rm phantomjs-2.1.1-linux-x86_64.tar.bz2 \
+  # Install PhantomJS binary to /usr/local/bin
+  && mv phantomjs-2.1.1-linux-x86_64/bin/phantomjs /usr/local/bin \
+  # Clean up extra (un-needed) PhantomJS files
+  && rm -rf phantomjs-2.1.1-linux-x86_64/
 
-# Add the apt-repository for the latest node.js
-RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
+# Specify a version of Node.js to download and install
+ENV NODEJS_VERSION=v10.15.3
 
-RUN apt-get update -qq && apt-get install -y build-essential libpq-dev nodejs
-RUN apt-get install build-essential chrpath libssl-dev libxft-dev -y && \
-  apt-get install libfreetype6 libfreetype6-dev -y && \
-  apt-get install libfontconfig1 libfontconfig1-dev -y && \
-  cd ~ && \
-  export PHANTOM_JS="phantomjs-2.1.1-linux-x86_64" && \
-  wget https://github.com/Medium/phantomjs/releases/download/v2.1.1/$PHANTOM_JS.tar.bz2 && \
-  tar xvjf $PHANTOM_JS.tar.bz2 && \
-  mv $PHANTOM_JS /usr/local/share && \
-  ln -sf /usr/local/share/$PHANTOM_JS/bin/phantomjs /usr/local/bin && \
-  apt-get install -y yarn
+# Download and extract Node.js from archive supplied by nodejs.org
+RUN curl -L https://nodejs.org/dist/$NODEJS_VERSION/node-$NODEJS_VERSION-linux-x64.tar.xz -o nodejs.tar.xz \
+  && tar xf nodejs.tar.xz \
+  # Clean up the Node.js archive
+  && rm nodejs.tar.xz
 
-# Clean up the apt cache
-RUN rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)


DeeDeeG:
I'm pretty happy with how this is set up now, after tweaking it a few times.

I'd like to merge this in about a week, if that's okay / if there are no objections.

To reiterate what this is for: basically this makes Docker builds run faster, and also makes them use less hard-disk space, and pull less data over the network.

DeeDeeG:
Not strictly in scope for this pull request, but I have a couple of other things in mind to help with faster/smaller builds:
 - Switch to the "minimal" base image at Travis, since we're doing everything inside our Docker container anyways. (We currently use the Ruby image at Travis, which is a larger base image to load, and also installs Ruby with RVM based on our `.ruby-version` file, and it also does some other Ruby-specific things.) Switching this could save maybe 30 seconds for Travis builds??
   - (Inspired by this: https://github.com/btyy77c/refugerestrooms/commit/393cf46603a01948af460a572a1cf2a770db8864)
   - (Documentation at Travis re: minimal/generic images: https://docs.travis-ci.com/user/languages/minimal-and-generic/) 
 - Switch our postgres Docker image to specifically use the postgres-alpine variant. (This is done in `docker-compose.yaml`.) This pulls a much smaller postgres image over the network, from what I can tell it (postgres) works exactly the same under Alpine vs Debian. (I believe the default postgres Docker image uses Debian.) And since Docker isn't used in production, we can try this out with no consequences to the actual production app. I think it would be worth trying, even though I admit I'm not sure how to troubleshoot Alpine Linux, the hope is that it "Just Works":tm:, and we don't have to think about it very much. Free network savings??
   - Also inspired by that branch: https://github.com/btyy77c/refugerestrooms/commits/dockerAlpine
   - Documentation section discussing alpine vs the default postgres image: https://hub.docker.com/_/postgres#image-variants (not super helpful IMO, but this is what I could find) (I think we have to just try it to find out if it works. I've tried it and it worked for me.)

DeeDeeG:
The only thing I'm on the fence about is `yarn install --frozen-lockfile` vs `yarn install --pure-lockfile`.

The difference seems to be, `--frozen-lockfile` will error out if packages from `package.json`, plus their dependencies, can't be installed in a way that matches the lockfile. with `--pure-lockfile`, Yarn doesn't care if there is a discrepancy, it simply installs stuff from `package.json` (plus dependencies) to ./node_modules and leaves the `yarn.lock` file alone.

I think `--frozen-lockfile` is better for CI. It's also more aggressive at enforcing we have a compatible package.json and yarn.lock, because error-ing out will stop Docker builds (`docker-compose build`). So for that reason, when noodling about with the package.json, it's actually much nicer/friendlier for the Dockerfile to have --pure-lockfile (which gives a free pass for any mismatch in the lockfile during development).

I almost don't get why --pure-lockfile exists, since it basically bypasses the whole functionality/purpose of the lockfile... But it does make tinkering and iterating potentially faster, so I guess that's why. But for CI I think we need --frozen-lockfile.

I am torn. As you can see from my commit, I leaned toward --frozen-lockfile.

DeeDeeG:
Went with `yarn install --pure-lockfile` to keep development as easy as possible. Allows tweaking the package.json without subsequent `docker-compose build` runs failing (erroring out).

This means the yarn command used in our Dockerfile is unchanged. I did add the word "`install`", because it's clearer/more readable IMO than without the sub-command.

PR #589, opened 2019-08-10
Dockerfile: Get latest Node.js within a major version

DeeDeeG:
# Context
- The recent pull request #586 changed the way Node.js is installed in the Docker container. In order to do a more lightweight installation, we directly download and extract Node.js as a tar archive. This replaced a complex installer script that did `apt install` steps, among other things, which were both slow to complete, and used much network/storage resources. As a part of the changes in that Pull Request, we specify an exact version of the form `node-v[major.minor.patch][...].tar.xz`
- The present pull request changes it so we specify only a major version number, and automatically fetch the latest minor/patch version of that major version number.

# Summary of Changes

- In the Dockerfile, switch from full explicit version number (e.g. node v10.16.0) to just specifying a major version (e.g. node v10) and get the latest minor/patch of that major version.

# Checklist

- [N/A] Added Unit Tests
- [x] CI Passes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,17 +15,20 @@ RUN curl -L https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.1.1-linu
   # Clean up extra (un-needed) PhantomJS files
   && rm -rf phantomjs-2.1.1-linux-x86_64/
 
-# Specify a version of Node.js to download and install
-ENV NODEJS_VERSION=v10.15.3
+# Specify a major version of Node.js to download and install
+ENV NODEJS_MAJOR_VERSION=10
 
 # Download and extract Node.js from archive supplied by nodejs.org
-RUN curl -L https://nodejs.org/dist/$NODEJS_VERSION/node-$NODEJS_VERSION-linux-x64.tar.xz -o nodejs.tar.xz \
+RUN curl -L https://nodejs.org/dist/latest-v$NODEJS_MAJOR_VERSION\.x/SHASUMS256.txt -O \
+  && ARCHIVE_FILENAME=$(grep -o "node-*.*.*-linux-x64.tar.xz" SHASUMS256.txt) \
+  && curl -L https://nodejs.org/dist/latest-v$NODEJS_MAJOR_VERSION.x/$ARCHIVE_FILENAME -o nodejs.tar.xz \
   && tar xf nodejs.tar.xz \
-  # Clean up the Node.js archive
-  && rm nodejs.tar.xz
+  && mv ./node-v*-linux-x64 /usr/local/nodejs \
+  # Clean up the Node.js archive and SHASUMS256.txt
+  && rm nodejs.tar.xz SHASUMS256.txt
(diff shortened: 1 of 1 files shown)


DeeDeeG:
The commit message explains this in much more detail:

> Dockerfile: Get latest Node.js in a major version
>
> --
>
> (Also installs Node.js in /usr/local/
> instead of installing Node.js in the root directory.)
> 
> With this updated script, we specify just a major version
> and the script picks the latest minor/patch version within that.
> 
> --
> 
> Nodejs.org does most of the work by maintaining the ["latest-v[MAJOR]"
> folders](https://nodejs.org/dist/); We only need to parse the ["SHASUMS256.txt"](https://nodejs.org/dist/latest-v10.x/SHASUMS256.txt) file from there,
> and pick the "linux-x64" variant, which works with our Docker setup.
> 
> At this point we can use the [known directory](https://nodejs.org/dist/latest-v10.x/) URL, plus the filename
> extracted from ["SHASUMS256.txt"](https://nodejs.org/dist/latest-v10.x/SHASUMS256.txt), and download with curl, or wget, etc.
> 
> (e.g. "curl -L https://nodejs.org/dist/latest-v10.x/node-v10.16.0-linux-x64.tar.xz -o nodejs.tar.xz")
> 
> --
> 
> There is no "latest-LTS" folder or similar, so automatically getting
> the latest LTS version would be more difficult.
> 
> We could search ["nodejs.org/dist/index.html"](https://nodejs.org/dist/index.html) for folders with
> the name "latest-[LETTERS-ONLY-STRING]", which would be the folders
> of all the LTS codenames. Among these, the one with
> the alphabetically last name is the latest LTS.
> 
> This would work at least until around 2040, when they may have
> to loop around and re-use some earlier letters (a, b, c, etc.)

DeeDeeG:
As an alternative to this PR, we can stop implementing our own "download Node.js" script, and instead rely on a popular node version manager. There are several options, such as [`nvm`](https://github.com/nvm-sh/nvm) and [`n`](https://github.com/tj/n).

(`n` is very lightweight, and I would lean toward using that. It is literally just a single-file bash script that takes up only 19KB. I had trouble getting `nvm` to work under Docker, so there's that also.)

Being fully-featured version managers, these can fetch the latest LTS, the latest major version, a specific minor or patch version, etc. They are much more flexible than my minimal install script (which takes up just a few lines of our Dockerfile.).

I notice `n` keeps two full (and extracted/uncompressed) copies of Node.js on the system, for no particular benefit to us, so I would probably want to delete the extra copy during our Dockerfile's install steps. But that would just be to keep our Docker setup from eating up disk space when folks are doing testing. (As has been pointed out in other PRs, we currently don't distribute pre-built Docker containers of Refuge Restrooms. It's just for local testing and Continuous Integration.)

On the other hand, I'm not sure we need the flexibility. We have gotten by so far with an install script that automatically gets the latest of a given major version.

DeeDeeG:
Planning to merge this today.

PR #595, opened 2019-10-02
switched sass gem

### LINKED ISSUE #507: Move from "sass" gem to "sassc"
## Context

See this blog post: http://sass.logdown.com/posts/7081811

And this issue over at sass-rails: rails/sass-rails#420

## Summary 

The ruby implementation of sass is designated as "deprecated", so it's in maintenance-only mode. They will be archiving the ruby-sass repo and not accepting issues or pull requests after March 29 2019. So it is going to eventually get stale.

## Solutions

We might want to consider switching to [`sassc-rails`](https://github.com/sass/sassc-rails) which uses the new C implementation of Sass.

(That said, the `sass-rails` folks [hope to be compatible with `sassc` under the hood at some point](https://github.com/rails/sass-rails/pull/349#issuecomment-300296084), so maybe this will be a moot point and we won't have to switch.)

btyy77c:
# Context
- Fixes #507
- Remove sass-rails gem and replaced it with the sassc-rails gem

# Summary of Changes

- Remove sass-rails gem from the Gemfile
- Added the sassc-rails gem to the Gemfile
- Updated Gemfile.lock

# Checklist

- [ ] Added Unit Tests
- [x] CI Passes

# Screenshots


DeeDeeG:
@btyy77c Thanks for this.

(Extra background info: the very latest `sass-rails` gem is just a thin wrapper for `sassc-rails`. I believe we now have that version of `sass-rails`, so we already pull in `sassc-rails` indirectly that way. However, `sass-rails` depends on a specific version range of the `sassc-rails` gem. It's neater for us to depend directly on `sassc-rails`, especially so that we keep getting updates as they are published, or can control the specific version of `sassc-rails`, should we need/want to do that.)

For details about the `sass-rails` --> wrapper for `sassc-rails` transition, see: https://github.com/rails/sass-rails/pull/424

We can keep #507 open until we update `activeadmin` to a version that doesn't pull in the old Ruby `sass` gem.

Gemfile
```diff
@@ -27,7 +27,7 @@ gem 'puma'
 gem 'rack-cors', :require => 'rack/cors'
 gem 'rack-jsonp'
 gem 'rakismet'
-gem 'sass-rails'
+gem 'sassc-rails', '~> 2.1'
```
DeeDeeG: I personally prefer we leave this without a version number. We can pull in the latest with `bundle update`, and CI/manual testing will make sure nothing breaks.

btyy77c: No problem.  I removed the version.

PR #620, opened 2020-04-10
Add a locale switcher

### LINKED ISSUE #428: Make a language switcher for the site

DeeDeeG:
# Context
- Finishes the work started in #585
- Fixes #428
- Adds a language switcher, rewires some of the routing and locale handling to support the locale switcher.

# Summary of Changes

- Add a default URL parameter, `:locale` that will render as `?locale=[your-locale-code-here]` in the URL. E.g. `/restrooms/new?locale=es` is the "Add a New Restroom" page, explicitly in Spanish.
- Enable optional routing to locale-prefixed versions of most URLs in the app. E.g. `/fr/restrooms/new` is the "Add a New Restroom" page, explicitly in French.
- Add locale-switching links to the footer, approach inspired by the footer at https://rubygems.org
- Most dynamically-generated links automagically use the locale prefixes (like `/pt-BR/`) now.
  - Exceptions appear to be... `/admin` (locale switcher and URL prefixing don't seem to work here.)
  - The site root/homepage `/` can work with locale prefixes, but the link in the header seems to prefer locale parameters `/?locale=[your-locale-code-here]`

# Checklist

- [ ] Added Unit Tests
- [ ] CI Passes

# Screenshots

### DIFF
--- a/app/assets/stylesheets/components/common.scss
+++ b/app/assets/stylesheets/components/common.scss
@@ -145,3 +145,7 @@ footer {
         color: #BE1E2D;
     }
 }
+
+.locale_link {
+    margin: 0px 5px;
+}
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,7 +10,7 @@ class ApplicationController < ActionController::Base
   end
 
    def set_locale
-     I18n.locale = http_accept_language.language_region_compatible_from(I18n.available_locales)
+     I18n.locale = params[:locale] || http_accept_language.language_region_compatible_from(I18n.available_locales)
    end
 
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -13,4 +13,8 @@ class ApplicationController < ActionController::Base
      I18n.locale = params[:locale] || http_accept_language.language_region_compatible_from(I18n.available_locales)
    end
 
+  def default_url_options
+    { locale: I18n.locale }
+  end
+
 end
--- a/app/views/layouts/_footer.html.haml
+++ b/app/views/layouts/_footer.html.haml
@@ -18,11 +18,18 @@
     = "\&copy; #{t('.copyleft')}  #{Date.today.year} #{t('.refuge-restrooms')}".html_safe
     %br/
     %br/
-    = link_to 'English', request.query_parameters.merge({:locale => 'en'})
-    = link_to 'Español', request.query_parameters.merge({:locale => 'es'})
-    = link_to 'Filipino/Tagalog', request.query_parameters.merge({:locale => 'fil'})
-    = link_to 'Français', request.query_parameters.merge({:locale => 'fr'})
-    = link_to 'हिन्दी', request.query_parameters.merge({:locale => 'hi'})
-    = link_to 'Italiano', request.query_parameters.merge({:locale => 'it'})
-    = link_to 'polski', request.query_parameters.merge({:locale => 'pl'})
-    = link_to 'Português do Brasil', request.query_parameters.merge({:locale => 'pt-BR'})
+    = link_to 'English', request.query_parameters.merge({:locale => 'en'}), class: "locale_link"
+    &bull;
+    = link_to 'Español', request.query_parameters.merge({:locale => 'es'}), class: "locale_link"
+    &bull;
+    = link_to 'Filipino/Tagalog', request.query_parameters.merge({:locale => 'fil'}), class: "locale_link"
+    &bull;
+    = link_to 'Français', request.query_parameters.merge({:locale => 'fr'}), class: "locale_link"
+    &bull;
+    = link_to 'हिन्दी', request.query_parameters.merge({:locale => 'hi'}), class: "locale_link"
+    &bull;
+    = link_to 'Italiano', request.query_parameters.merge({:locale => 'it'}), class: "locale_link"
+    &bull;
+    = link_to 'polski', request.query_parameters.merge({:locale => 'pl'}), class: "locale_link"
+    &bull;
+    = link_to 'Português do Brasil', request.query_parameters.merge({:locale => 'pt-BR'}), class: "locale_link"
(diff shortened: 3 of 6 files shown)


DeeDeeG:
If anyone knows how to make this test passing again, I'd appreciate it a lot! :heart:

```ruby
Failures:

  1) the contact process should show a generic contact when contact is not from restroom form
     Failure/Error: visit restroom_path restroom
     
     ActionController::UrlGenerationError:
       No route matches {:action=>"show", :controller=>"restrooms", :locale=>#<Restroom id: 232, name: "Mission Creek Cafe", street: "1800 Market St", city: "San Francisco", state: "CA", accessible: false, unisex: false, directions: "Direction", comment: "Comment", latitude: nil, longitude: nil, created_at: "2020-04-11 00:04:34", updated_at: "2020-04-11 00:04:34", downvote: 11, upvote: 22, country: "US", changing_table: false, edit_id: 232, approved: true>}, missing required keys: [:id]
     # /usr/local/bundle/gems/actionpack-5.2.4.2/lib/action_dispatch/journey/formatter.rb:57:in `generate'
     # /usr/local/bundle/gems/actionpack-5.2.4.2/lib/action_dispatch/routing/route_set.rb:744:in `generate'
     # /usr/local/bundle/gems/actionpack-5.2.4.2/lib/action_dispatch/routing/route_set.rb:775:in `generate'
     # /usr/local/bundle/gems/actionpack-5.2.4.2/lib/action_dispatch/routing/route_set.rb:822:in `url_for'
     # /usr/local/bundle/gems/actionpack-5.2.4.2/lib/action_dispatch/routing/route_set.rb:273:in `call'
     # /usr/local/bundle/gems/actionpack-5.2.4.2/lib/action_dispatch/routing/route_set.rb:214:in `call'
     # /usr/local/bundle/gems/actionpack-5.2.4.2/lib/action_dispatch/routing/route_set.rb:331:in `block (2 levels) in define_url_helper'
     # ./spec/features/contacts_spec.rb:7:in `block (2 levels) in <top (required)>'

Finished in 40.73 seconds (files took 1.95 seconds to load)
62 examples, 1 failure

Failed examples:

rspec ./spec/features/contacts_spec.rb:4 # the contact process should show a generic contact when contact is not from restroom form
```

It seems like when `visit restroom_path restroom` is invoked here, a garbled `:locale` value gets piped in that has way too much unrelated data in it, and so it isn't valid as a locale. And the app gets rather unhappy about it.

---

Viewing restroom listings, and using the contacts page both work locally in Docker.

I think the test itself is broken here, and the app isn't misbehaving (other than not matching up well with the test anymore).

Since this test is related to using the "contact" page from a restroom listing, and that functionality was removed in https://github.com/RefugeRestrooms/refugerestrooms/pull/487, I'm somewhat tempted to just disable or greatly simplify this test.

mi-wood:
That test still sort of acts as a regression tests and tests navigation to `Contact`. It's implicitly filling in the parameters of `restroom_path` and setting `:locale` to the `restroom`. You can explicitly set it as `restroom_path(:id => restroom.id)`.

https://stackoverflow.com/questions/31224170/no-route-matches-action-show

DeeDeeG:
Thanks, @mi-wood, tests are passing now!

I think this is ready to merge.

DeeDeeG:
Gong to put these changes on staging for a bit, but it seems worth another deploy if they behave well.

DeeDeeG:
Turns out this does break some of the locale-prefix-less routes, like `/about`, `/business_info`, `/contact`, `/signs` and the hidden `/text`. It doesn't break `/restrooms/new`, `/restrooms/[id]`, or `/api/docs` though.

Either I need to figure out a fix (or someone can explain one to me/commit it themselves) or perhaps revert this PR for now in `develop` branch.

I find routing in Rails kind of tricky.

DeeDeeG:
Reverted for now, as this wasn't working as intended.

This was working with just the `?locale=[your-locale-here]` URL parameters, but was still kinda broken with the `/[your-locale-here]/` URL prefixes.

I would really like to get the URL prefixes working, but I need to read up more on the Rails router. (If anyone knows how to do this, feel free to step in!)

Also, there are hard-coded HTML anchor (`<a>`) tags from about to business_info, and vice-versa, but these are inside Internationalization (I18n) strings. I have to look up the interaction between haml and I18n here to properly translate the dynamic link with `link_to` and `[page-here]_path` helpers.

PR #628, opened 2020-06-24
Add listeners to close infowindow when another is opened

### LINKED ISSUE #627: Map selection does not close when another is selected
## Scope / difficulty

Difficulty: easy-medium
Files to change: `maps.js.erb`

## Impact

- Improve user experience when navigating the map view. Currently, bathroom info can "build-up" on the map, like so:

- This project will auto-close bathroom markers when a new one is selected, and close the markers when the user clicks "out" of the resource to the map. 

## Rationale

- This change will prevent selections from overlapping and gathering on the map view. 
- It's not the most ideal experience to have to click a small `x` to close the info window every time. 
- Bathroom resources should open and close as a new marker is selected. 
- These changes will more-closely mimic user interactions with other maps, for example:

## Proposal

- Add logic to close opened-bathrooms when a new bathroom is selected. 
- Add an on-click listener to the map when the user clicks "out" of the marker to close the opened marker. 

These will prevent bathrooms from crowding the map and overlapping each other. 

## How to actually do this:

- The map view code lives in [`maps.js.erb`](https://github.com/RefugeRestrooms/refugerestrooms/blob/develop/app/javascript/packs/lib/maps.js.erb). 
- To close a marker when another marker is selected:
   -  Add code to close any opened `infoWindows` when a marker is clicked. Probably add the logic in the existing on-click listener [here](https://github.com/RefugeRestrooms/refugerestrooms/blob/develop/app/javascript/packs/lib/maps.js.erb#L83).  
   - Since each `marker` has its own relative `infoWindow`, as opposed to a global `infoWindow`, I think the most recently opened window should be captured in a global variable as suggested [here](https://stackoverflow.com/questions/12621274/close-infowindow-when-another-marker-is-clicked), and then closed as the new marker is selected.
- To close opened markers when the user clicks "out" of the bathroom (i.e. clicks out to the map), an on-click listener should be added to the map as described [here](https://stackoverflow.com/questions/10022873/closing-info-windows-in-google-maps-by-clicking-the-map). 
   - I think this listener should be added within the [`placeMarker()` function](https://github.com/RefugeRestrooms/refugerestrooms/blob/develop/app/javascript/packs/lib/maps.js.erb#L56), so the correct `infoWindow` is closed.

piercebb:
# Context
- Fixes #627 
- Select one bathroom at a time in the Map View
- Close windows when clicking "out" to the map

# Summary of Changes

- Added on-click listener to close an open `infoWindow` when the map is selected
- Added a global variable to capture the most recently opened `infoWindow`
- Added logic to close an opened `infoWindow` when another marker (bathroom) is selected

# Checklist

- [ ] Added Unit Tests
- [ ] CI Passes

# Screenshots

### DIFF
--- a/app/javascript/packs/lib/maps.js.erb
+++ b/app/javascript/packs/lib/maps.js.erb
@@ -8,6 +8,7 @@ const circleMarkerImage = "<%= helpers.image_path('marker-circle.png') %>";
 const showMarkerImage = "<%= helpers.image_path('show-icon.png') %>";
 
 let map;
+let lastOpenedInfoWindow;
 
 function initMap(x, y, image, draggable, callback){
   image = typeof image !== 'undefined' ? image : currentLocationImage;
@@ -81,8 +82,20 @@ function placeMarker(lat, lng, content, number){
   });
 
   marker.addListener('click', function() {
+    closeLastOpenedInfoWindow();
     infoWindow.open(map, marker);
+    lastOpenedInfoWindow = infoWindow;
   });
+
+  google.maps.event.addListener(map, 'click', function(event) {
+    infoWindow.close();
+  });
+}
+
+function closeLastOpenedInfoWindow() {
+    if (lastOpenedInfoWindow) {
+        lastOpenedInfoWindow.close();
+    }
 }
 
 function generateContent(data){


DeeDeeG:
The code change looks very straightforward, easy to review.

CI passed as well. I'd like to poke around with this in Docker on my machine and just give it time in case something jumps out as needing a change, but otherwise, I don't see why not to merge this.

(@stardust66 you are pretty good with javascript! If you happen to want to take a look at this, your feedback would be appreciated. It's a  very small change though, so don't feel bad if you are busy. I can test it myself, but multiple reviewers is always a great thing to strive for IMO.)

Once again, thank you for the contribution @piercebb, and thank you for filling out the PR template so fully. It really illustrates what is going on with this change.  ❤️ Very easy to review.

---

Tests:

If it was a bigger change (and in general) it's good to have tests for every little feature, but this is more of a "polish/nice-to-have" thing and it's okay without a test. If you are up for it, technically we do like to have tests for as much of the codebase as possible. It would be something like "When I click on the map, an already-open restroom info popup is closed". And simulate a click somewhere on the map with Capybara. We have [some docs about that](https://github.com/RefugeRestrooms/refugerestrooms/wiki/What-is-Capybara%3F-What-is-PhantomJS%3F-What-is-Poltergeist%3F) in our Wiki.

I am comfortable in this case with not requiring a test, because I think the test would be harder to code than the feature itself, and as mentioned above, it's a nice-to-have feature, not a core feature.

Let me know if you are working on a test and then I will leave this open to make sure you have time to get that working. If I don't hear back about that, it's fine and I'll move ahead without expecting a test to be added for this. Truly fine either way. Could be a good learning experience for some of our contributors, so that's a reason I suggest it as well.

PR #626, opened 2020-06-23
Use `VAR=VAL` syntax for environment variables

piercebb:
# Context
- Fixes #624 
- These changes unblock [a common docker-compile error](https://github.com/RefugeRestrooms/refugerestrooms/wiki/Troubleshooting-Docker#error-in-file-env-environment-variable-name--may-not-contain-whitespace). 
- Changes now use Docker's recommended syntax for [Declaring default environment variables](https://docs.docker.com/compose/env-file/#compose-file-and-cli-variables) in `.env` files. 

# Summary of Changes

- Use `VAR=VAL` syntax for default environment variables
- Remove `export VAR=` syntax

# Checklist

- [ ] Tested Mobile Responsiveness
- [ ] Added Unit Tests
- [ ] CI Passes
- [ ] Deploys to Heroku on test Correctly (Maintainers will handle)
- [ ] Added Documentation (Service and Code when required)

PR #632, opened 2020-07-10
Delete .env file to investigate it's impact on project (#624)

### LINKED ISSUE #624: .env file variable syntax is not compatible with latest docker-compose version 1.25.5
## The issue
During project setup after cloning the project and installing docker, running `docker-compose` generated an error:
```
pierce@peridot refugerestrooms % docker-compose build
ERROR: In file ./.env: environment variable name 'export DEVISE_SECRET_KEY' may not contain whitespace.
```
Others have run into[ similar issues](https://github.com/docker/compose/issues/6511) using `export MY_VAR=MY_VAL` in `.env` files, which are [compatible with `docker-compose` version 1.23.1 ](https://github.com/docker/compose/issues/6511#issuecomment-509873061). 

## Scope / difficulty
I anticipate this will be a straight-forward fix: updating the syntax of the variables as [outlined in the docker documentation](https://docs.docker.com/compose/env-file/#compose-file-and-cli-variables). 

## Impact
Updating `.env` will un-block setting up the project and support developers as they set up their development environment. 

## Rationale
I ran into this issue during project setup and I anticipate others may get blocked as well.

## Proposal
Update `.env` variables so that they are compatible with the latest docker-compose version 1.25.5. 

## How to actually do this:
1. Edit the `.env` file
2. Change variable declaration to be `VAR=VAL` format instead of `export VAR=VAL`. 

See Docker's docs on declaring default environment variables [here](https://docs.docker.com/compose/env-file/#compose-file-and-cli-variables).

Joranhezon:
Context
- Fixes #624 
- Delete .env file to investigate it's impact on project.

# Summary of Changes

- Delete .env file.

# Checklist

- [ ] Added Unit Tests
- [ ] CI Passes

# Screenshots

## After
Screenshot from 2020-07-10 10-53-10

## Before
Screenshot from 2020-07-10 11-01-10


Joranhezon:
It seems that deleting the file won't solve the problem, since it breaks the app.

DeeDeeG:
@Joranhezon thanks for trying this. I will look at what changed and try to understand.

The other maintainers have set up this part of the app, so I am a bit behind on knowledge of this area.

(As always, @tkwidmer @mi-wood if you know the answer to this one, and you have a free bit of time, feel free to elucidate!)

Will keep trying to get to the bottom of it. :+1:

DeeDeeG:
There is also #626, which may be a real solution. But by trying something more dramatic, hopefully the errors will shed light on the technical details of why we need these variables.

DeeDeeG:
Continuous integration tests passed, https://github.com/RefugeRestrooms/refugerestrooms/pull/632/checks?check_run_id=858237466, so I think if there are problems they are in areas that we don't have tests for.

DeeDeeG:
I have found that without this `.env` file, the `RECAPTCHA_SITE_KEY` and `RECAPTCHA_SECRET_KEY` environment variables are unset, and recaptcha no-longer works in local testing with Docker.

(Apparently these keys are set to work on `localhost` only, so having them out in the open is not a concern, with regard to someone potentially abusing these keys.)

Given that there is some usefulness to this file, I am planning to keep it for the time being. Thanks again for taking an interest and helping to look into this. As a late-comer to this project, I am always learning more about its codebase. This helped me understand why we need this file. 👍

DeeDeeG:
Closing, as it turns out we want to keep this file after all.

PR #641, opened 2020-10-07
Housekeeping/Upgrade: Puma

brunoocasali:
# Context
- Puma released a new stable version (5.0) and in the company I work for we already using it, and we got some nice improvements. For this PR I've upgraded to version 5.0 and added clustered mode configuration (which removes latency/request queuing times)

Related: https://www.speedshop.co/2017/10/12/appserver.html

# Summary of Changes

- Upgrade puma gem to 5.x
- Added clustered mode to serve more requests and reduce latency

# Checklist

- [ ] Tested Mobile Responsiveness
- [ ] Added Unit Tests
- [ ] CI Passes
- [ ] Deploys to Heroku on test Correctly (Maintainers will handle)
- [ ] Added Documentation (Service and Code when required)

PR #647, opened 2020-10-14
Enhancement/i18n thread safe

brunoocasali:
# Context
This code makes any request free of locale leaking problems.

According to the rails guides: https://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests
Using `I18n.locale` is not a recommended way to set current locale for a request, using the `with_locale` is the best option!

Btw, searching more about the `http_accept_language` I've found this https://github.com/iain/http_accept_language/issues/68

# Summary of Changes

- Remove `I18n.locale`
- Create shared_examples to test shared behaviour between specs

# Checklist

- [x] Added Unit Tests
- [x] CI Passes

### DIFF
--- a/spec/controllers/pages_controller_spec.rb
+++ b/spec/controllers/pages_controller_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-describe PagesController, type: :controller  do
+describe PagesController, type: :controller do
   it_behaves_like 'localized request', :index
 
   it "#index" do
--- a/spec/controllers/pages_controller_spec.rb
+++ b/spec/controllers/pages_controller_spec.rb
@@ -1,6 +1,8 @@
 require 'spec_helper'
 
-describe PagesController, type: :controller do
+describe PagesController, type: :controller  do
+  it_behaves_like 'localized request', :index
+
   it "#index" do
     get :index
     assert_response :success
--- a/spec/support/shared_examples/localized_request.rb
+++ b/spec/support/shared_examples/localized_request.rb
@@ -1,0 +1,9 @@
+shared_examples_for 'localized request' do |action|
+  it 'calls I18n.with_locale in requests' do
+    allow(I18n).to receive(:with_locale)
+
+    get action
+
+    expect(I18n).to have_received(:with_locale).with(:en)
+  end
+end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,14 +2,18 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
+
+  around_action :switch_locale
   before_action :mobile_filter_header
-  before_action :set_locale
 
   def mobile_filter_header
     @mobile = true
   end
 
-  def set_locale
-    I18n.locale = http_accept_language.language_region_compatible_from(I18n.available_locales)
+  def switch_locale(&action)
+    locale = http_accept_language.language_region_compatible_from(I18n.available_locales)
+    locale ||= I18n.default_locale
+
+    I18n.with_locale(locale, &action)
   end
 end
--- a/spec/controllers/restrooms_controller_spec.rb
+++ b/spec/controllers/restrooms_controller_spec.rb
@@ -1,6 +1,8 @@
 require 'spec_helper'
 
 describe RestroomsController, type: :controller do
+  it_behaves_like 'localized request', :index
+
   it "#index" do
     get :index
     assert_response :success


brunoocasali:
This PR is related to https://github.com/RefugeRestrooms/refugerestrooms/pull/641 :)

DeeDeeG:
cc @mi-wood again, or @tkwidmer. This seems like a good PR, making sure our internationalization (I18n) is ready for multi-threaded Puma. It even adds tests!

I'm personally not very good with Ruby, so I don't think my review would be super meaningful here. If anyone good with Ruby can give this a second pair of eyes, that would be great. Or at least give it the thumbs up so we can approve it, so the author gets credit for this PR toward their `Hacktoberfest` "5 good pull requests" goals.

brunoocasali:
@DeeDeeG @tkwidmer rebased with latest version of develop branch, can I do something to help this get merged? :)

DeeDeeG:
@brunoocasali thank you for rebasing!

I have simply been backed up with the number of pull requests we got this month, plus other life things.

(I'm trying to work through the remaining pull requests. The fewer of them that are left, the easier reviewing each one becomes, due to not having to worry which one will have merge conflicts, or which one will cause the next one to be harder to test.)

I am going to ensure that all the helpful Pull Requests we have received are either "Approved" or merged by the end of the month. I believe all of your Pull Requests are looking good, and we will be approving and eventually merging them. Thanks for sticking with this as I work through the backlog.

spec/controllers/pages_controller_spec.rb
```diff
@@ -1,6 +1,8 @@
 require 'spec_helper'
 
-describe PagesController, type: :controller do
+describe PagesController, type: :controller  do
```
DeeDeeG, line 3: Hi, @brunoocasali, there is an extra space here. Rubocop is flagging this.

```suggestion
describe PagesController, type: :controller do
```

DeeDeeG:
I do recall the documentation for I18n saying this is the correct way. If there are no issues in manual testing, then I am inclined to merge this.

DeeDeeG:
Okay, this works.

I tested by going to about:preferences in Firefox, updating the "preferred language for displaying pages", then visiting links/refreshing the page at https://refuge-deedeeg-test.herokuapp.com. My "preferred language" was always respected accurately.

(As I understand it, this will only really make a difference if/when we enable multiple workers for Puma some time in the future. Note to other maintainers/readers: see #641 for discussion about multiple workers in Puma. May or may not be merged soon, depending on how review/discussion goes.)

I then compared page load times between:
- https://refuge-deedeeg-test.herokuapp.com (this branch)
- https://staging.refugerestrooms.org (`develop` branch)

I didn't notice a significant difference; either one would be faster or slower seemingly at random. So this pull request does not appear to have any adverse impact on performance.

I'll watch this change on the staging instance, after merging, and hopefully it will continue to behave well as intended in the coming days. In my opinion, this is ready to merge.

spec/support/shared_examples/localized_request.rb
```diff
@@ -0,0 +1,9 @@
+shared_examples_for 'localized request' do |action|
+  it 'calls I18n.with_locale in requests' do
+    allow(I18n).to receive(:with_locale)
+
+    get action
+
+    expect(I18n).to have_received(:with_locale).with(:en)
+  end
+end
```
DeeDeeG, line 9: @brunoocasali would it be very difficult to also test loading a page with another locale? So far our test quite runs strictly against the site in English. I would love to have some test for a locale other than English, even if it was a very basic or simple test.

If you could do that, it would be much appreciated. If it's a lot of work, let me know, and I'll still accept this pull request without it.

Thank you, and best regards.

\- DeeDeeg

DeeDeeG (approved):

spec/support/shared_examples/localized_request.rb
brunoocasali: Hi @DeeDeeG, thanks for your review! 

Don't you think that way we could start testing some code that are already tested by the framework?

The test in this PR just guarantees that every request will contain some kind of language definition. 
When we add a multi-language switcher like a querystring called "lang" for example we could add another spec here, to check.

if the url contains the "lang" querystring all the messages will be translated properly.

brunoocasali:
> I didn't notice a significant difference; either one would be faster or slower seemingly at random. So this pull request does not appear to have any adverse impact on performance

Yeah, this change will not improve the performance or impact in a negative way either. This just makes sure that there is no way someone will request a page in "pt-BR" and get the answer in "fr" because someone requested nanosecods before :)

spec/support/shared_examples/localized_request.rb
DeeDeeG: It should be fairly easy to set the locale within a spec's code.

But I can see how that would be unrelated to this pull request. Perhaps when I re-add the language switcher feature, I will `@mention` you to see if you might be interested in adding some tests. No obligation, but certainly it would be appreciated if you wanted to help with that.

Thanks for this, I'm going to merge now.

brunoocasali: @DeeDeeG of course! just mention me and I see what can I do ;)

PR #654, opened 2020-10-29
Enable puma clustered

brunoocasali:
Based on: https://github.com/RefugeRestrooms/refugerestrooms/pull/641

This PR just enable the puma clustered mode as discussed in https://github.com/RefugeRestrooms/refugerestrooms/pull/641#discussion_r513603993

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -21,14 +21,7 @@ environment ENV.fetch("RAILS_ENV") { "development" }
 # Workers do not work on JRuby or Windows (both of which do not support
 # processes).
 #
-# workers ENV.fetch("WEB_CONCURRENCY") { 2 }
-
-# Use the `preload_app!` method when specifying a `workers` number.
-# This directive tells Puma to first boot the application and load code
-# before forking the application. This takes advantage of Copy On Write
-# process behavior so workers use less memory.
-#
-# preload_app!
+workers ENV.fetch("WEB_CONCURRENCY") { 3 }
 
 # Allow puma to be restarted by `rails restart` command.
 plugin :tmp_restart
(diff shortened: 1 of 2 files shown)


config/puma.rb
```diff
@@ -21,14 +21,7 @@
 # Workers do not work on JRuby or Windows (both of which do not support
 # processes).
 #
-# workers ENV.fetch("WEB_CONCURRENCY") { 2 }
-
-# Use the `preload_app!` method when specifying a `workers` number.
-# This directive tells Puma to first boot the application and load code
-# before forking the application. This takes advantage of Copy On Write
-# process behavior so workers use less memory.
```
brunoocasali, line 29: In puma 5 preloading app is the default behaviour

config/puma.rb
```diff
@@ -21,14 +21,7 @@
...
-#
-# preload_app!
+workers ENV.fetch("WEB_CONCURRENCY") { 3 }
```
brunoocasali, line 24: @DeeDeeG if you could share with me some data about the workload in production would be nice, and I could help to improve this setup.

DeeDeeG: @brunoocasali hmm, I don't have access to that data myself.

cc @mi-wood @tkwidmer do you have data about our traffic/resource usage on Heroku?

This pull request is to look at enabling multiple workers/concurrency in Puma. It would be useful to know what our performance needs are. If you have anything relevant to share from the Heroku dashboards, that would be great.

Thanks.

brunoocasali: If you people use any APM will be nice to see them too, like new relic, scout apm or others..

DeeDeeG:
Merging this to `develop` branch... this makes the change live on https://staging.refugerestrooms.org, and queues the change for eventually being published to the main site at https://refugerestrooms.org.

I might revert this just before publishing a new release, then revert the revert again so it stays live on staging. That is, I'd like to try this out and be sure it works well for a while before putting this change on the main site. But I'm happy to begin trying it out now!

Thanks again!

PR #656, opened 2020-11-04
Add a locale switcher, second try

### LINKED ISSUE #428: Make a language switcher for the site

DeeDeeG:
# Context
- Finishes the work started in #585 and continued in #620
- Fixes #428
- Adds a language switcher, rewires some of the routing and locale handling to support the locale switcher.

# Summary of Changes

- Add a default URL parameter, `locale` that will appear as `?locale=[your-locale-code-here]` in the URL. For example: `/restrooms/new?locale=es` is the "Add a New Restroom" page, explicitly in Spanish.
  - Dynamically generated links, such as the ones made with Rails' helpers (for example: `link_to`), automagically use the URL parameter `?locale=[some_locale_here]` now.
- Use dynamically-generated links throughout the site, replace hard-coded links with dynamically-generated ones.
- Add locale-switching links to the footer, approach inspired by the footer at https://rubygems.org
- ~~Enable optional routing to locale-prefixed versions of most URLs in the app. For example: `/fr/restrooms/new` is the "Add a New Restroom" page, explicitly in French.~~ (Disabled for now, as this wasn't working as intended.)

# Checklist

- [ ] Added Unit Tests
- [x] CI Passes

# Screenshots

## Before
Refuge Restrooms website footer with GitHub, Twitter, Facebook and "Contact us" links, GitHub blurb with link, Patreon blurb with link, and copyleft notice

## After
Refuge Restrooms footer, with all of the above, and now with links to all locales at the bottom (English, Español, Filipino/Tagalog, Français, हिन्दी, Italiano, polski, and Português do Brasil)


DeeDeeG:
Hi @brunoocasali, I hope you are doing well.

I've been working on finishing the locale switcher feature for Refuge Restrooms. I think it's ready now!

If you are interested in helping with adding tests for locales other than English, I think this could be a reasonable place to do that. It would, of course, be much appreciated.

Best Regards,

\- DeeDeeG

brunoocasali (approved):
🥇 good work @DeeDeeG 

After these changes get merged, I can submit a PR to add tests to it!

DeeDeeG:
@brunoocasali I added the changes you suggested, matching the Rubocop-linted style.

During testing, I found out that there is a problem still. Might you be able to help troubleshoot this?

When searching for restrooms using the search bar, the locale parameter is not preserved. It appears to me that the search results page is generated only with the following parameters (with the example search "San Francisco").

`utf8=✓&lat=37.7749295&long=-122.4194155&search=san+francisco`

- `utf8`
- `lat`
- `long`
- `search`

But no `locale` parameter.

It's not immediately clear to me how these parameters are generated, as I'm not familiar with that part of the codebase. (I suspect it might be part of Refuge Restrooms' JavaScript code, but I'm not sure.) If I knew where these were generated, then perhaps it would be easier to ensure the `locale` parameter was preserved there, too, if set.

brunoocasali:
@DeeDeeG I could help you with this tomorrow, okay? :)

DeeDeeG:
That would be great.

I'm getting more confident that it's in `app/javascript/packs/lib/` --> `geocoder.js` or `maps.js.erb`. Not totally sure, though.

brunoocasali:
@DeeDeeG checkout the #657 ;)

PR #675, opened 2022-09-18
update to new maps public key

mi-wood:
# Context
- Public Google Maps API key has been having issues. Created a new one with new limits.


spec/features/restrooms_spec.rb
```diff
@@ -100,7 +101,7 @@
 
       click_on "Preview"
 
-      expect(page).to have_css("div#mapArea", visible: :visible)
+      page.has_css?(".nearby-container .listItem", visible: :visible)
```
DeeDeeG, line 104: I think the original CSS selector was correct here. (New one was probably copy-pasted from a nearby spec by accident?)

Looking to revert this back to the previous CSS selector soon if that's okay.

(Already said it elsewhere, but thanks again for fixing all this, and the dependency updates are appreciated as well.)

PR #681, opened 2023-05-24
Replace kaminari with pagy

### LINKED ISSUE #650: Replace kaminari with pagy
## Scope / difficulty

**[medium]** Replace **kaminari** with **pagy** in grape API.

## Impact

Actually the kaminari implementation is easy to plug and play, but it have some disadvantages because it generates tons of objects in memory. The pagy implementation from the other side is small and consumes less memory. You can learn more about it here:

- https://github.com/ddnexus/pagy
- https://www.imaginarycloud.com/blog/paginating-ruby-on-rails-apps-with-pagy/#:~:text=Pagy%20is%20using%20far%20less,still%207x%20more%20than%20Pagy.

I think we could improve a lot the api timings with this change

amadayami:
# Context
- Fixes #650
- Replaces [kaminari](https://github.com/kaminari/kaminari) pagination with [pagy](https://ddnexus.github.io/pagy/) pagination

# Checklist

- [ ] Added Unit Tests
- [x] CI Passes

# Screenshots

## Before

Screen Shot 2023-05-23 at 20 21 01

## After

Screen Shot 2023-05-23 at 20 19 58


app/controllers/api/v1/restrooms.rb
```diff
@@ -11,6 +10,10 @@ class Restrooms < Grape::API
       resource :restrooms do
         desc "Get all restroom records ordered by date descending."
         params do
+          use :pagy,
+              items_param: :per_page,
+              items: 10,
+              max_items: 100
```
Joranhezon: I noticed these lines are being reused in each desc, would it be possible to set this param for all of them outside of each example? Kinda like line 5 on the old file.

Something like `use :page, items_param :per_page, items: 10, max_items: 100` on line 5 ofthis file.

app/views/restrooms/index.html.haml
```diff
@@ -26,20 +26,20 @@
     .col-sm-12.headroom
       %ul{:class => "restrooms-list", :id => "list"}
         = render @restrooms
-        = paginate @restrooms
+        != pagy_nav @pagy
```
Joranhezon, line 29: Is there a comparison here we could make to keep the `=` instead of `!=`?

Joranhezon (commented):
Looks good to me! Thanks for working on this. (: 

I have some suggestions, but nothing major. ✌️

app/views/restrooms/index.html.haml
amadayami: Yes! It can be rewritten in this way to maintain the `=` that was there previously. This is functionally the same as using `!=`.

```suggestion
        = pagy_nav(@pagy).html_safe
```

Originally, the `!=` [evaluates the Ruby without ever sanitizing the HTML](https://haml.info/docs/yardoc/file.REFERENCE.html#unescaping_html). Rewritten, the `=` is maintained but it's [assumed to be safe](https://api.rubyonrails.org/v6.1.7.3/classes/String.html#method-i-html_safe) by marking it with `html_safe`. It has to be done in this format because pagy is an "[agnostic pagination gem it does not use the specific html_safe rails helper on its output](https://ddnexus.github.io/pagy/docs/how-to/#ignore-brakeman-unescapedoutputs-false-positives-warnings)".

app/controllers/api/v1/restrooms.rb
amadayami: Yes, I have changed it so that those parameters are written as a class constant for the descs to reference instead and have [committed that change](https://github.com/RefugeRestrooms/refugerestrooms/pull/681/commits/bf7126f4754375b161daf8f4ad4964fc56718701). Grape Pagy does not appear to have anything equivalent to the old line 5, so this would be the best way to extract those parameters.

amadayami:
The check failed after the CI errored in the build phase. Would someone be able to rerun the check? Thanks 😄

DeeDeeG:
Hi @amadayami,

I resolved the main issue (where tests were not getting to run at all) on our main `develop` branch. With that change on this branch, I assume _all but one_ test would pass, since it seems to rely on the Google Maps API working, and there's a bit of a quota or billing issue with that at the moment that's affecting the main site as well.

(Other maintainers have access to the Google API dashboards, and I don't, so I can't do anything about that personally, though I have pinged the other maintainers.)

Anyhow, **if you'd like to rebase on, or merge `develop` branch to this one,** that should get CI _mostly_ working again. (I can rebase or merge it for you, if you rather a maintainer take care of it.) And if it's the same one failure as we're seeing on `develop` branch right now, then I'd know it isn't this particular PR's fault.

amadayami:
Hi @DeeDeeG 
I have merged my branch with the updated develop branch and had no merge conflicts. I addressed the rubocop issues, but the one error you mentioned was still popping up (./spec/features/restrooms_spec.rb:45).

DeeDeeG (approved):
Thank you @amadayami for the code contribution and @Joranhezon for the help with review!

I tested it locally and it appears to work as intended. Tests pass other than the one with the known issue at the moment unrelated to this PR.

Looks good to me. Merging! (Thanks again!)

PR #685, opened 2023-11-26
Update Dependencies

DeeDeeG:
# Summary of Changes

- Update several NodeJS packages and Ruby Gems in our dependency specifications files

# Checklist

- [x] Tested Mobile Responsiveness / Manually tested the app basically works
- [ ] Added Unit Tests
- [ ] CI Passes


DeeDeeG:
This is working for me on my local machine.

I did `docker compose up` and visited `localhost:3000`... Searching for "San Francisco" does work, and the results from the testing placeholder DB entries do show in the results.

Furthermore, all tests pass but the "search from the splash page" test... Not sure why that isn't working in the specs, but it works in manual testing.

**I think the test "failure" is a _test-only_ failure.** I'm intending to merge this as-is, because I don't really have the time/energy at the moment to debug some weird test-only failure when it's actually working IRL. I'll be sure to check `staging.refugerestrooms.org` and confirm it works after merging this. Having the app actually working is more important to me than this frequently-problematic test passing.

**If at all possible, this test should be re-written so the results from Google Maps (or whatever maps/location provider) are stubbed/fake, so that everything about the test is fully local, and doesn't require live results from a real cloud service.** The test has been relatively error-prone in the past, so improving it would be a help for (hopefully) not having to pause to debug the tests themselves nearly as often.

PR #688, opened 2023-12-07
Update to NodeJS v20.x (using a workaround)

### LINKED ISSUE #682: Update app to run on Node 18 or newer (NodeJS 16 has reached End of Support)
See:

- https://github.com/nodejs/release#release-schedule
- https://nodejs.org/en/blog/announcements/nodejs16-eol

NodeJS 16 has reached End of Support as of 11th September 2023.

We should update some things to get the app running on newer NodeJS, which for still-supported versions would be Node 18 or newer.

I tried to do so during https://github.com/RefugeRestrooms/refugerestrooms/pull/678, but ran into issues with our older version of Webpack not working with Node 18, and it not being obvious to me personally how to properly migrate to newer Webpack either.

If anyone can help get the app running and functional (passing our CI tests, etc.), that would be much appreciated! I'm marking this issue as "Help Wanted" just in case it helps draw attention from some helpful folks to get us past this tech stack upgrade. Thanks in advance for any help with this.

Best Regards,
\- DeeDeeG

DeeDeeG:
# Context
- Sort-of resolves #682, with a workaround.
  - (Proper compatibility with NodeJS 17+, without resorting to a workaround, is still To-Do.)
- Use the workaround flag `--openssl-legacy-provider`, via the `NODE_OPTIONS` env var, so we can bump to newer NodeJS (Node v20.x, specifically).
- Note: NodeJS 18 was tested and working as well, but bumping to NodeJS 20 is more desirable, since it will be supported for longer.

# Summary of Changes

- Set the `NODE_OPTIONS=--openssl-legacy-provider` env var for development (`Dockerfile`) and production (`app.json`).
- Bump to NodeJS 20 in development (`Dockerfile`) and production (`engines` field of `package.json`).

# Checklist

- [x] Tested Mobile Responsiveness (and that app generally works in manual testing)
- [ ] Added Unit Tests
- [ ] CI Passes


DeeDeeG:
Other than the one [known failure from before](https://github.com/RefugeRestrooms/refugerestrooms/pull/685#issuecomment-1826911109), CI is passing.

Manual testing worked fine, even for the functionality tested by the "failing" test case. Searching from the splash page (AKA homepage) "just works". _Still_ not sure why that's broken in testing, since it does work in manual testing. But I'm inclined to ignore that apparently test-only failure if it allows getting us on newer dependencies/runtime versions.

And, for what it's worth, upgrading all outdated ruby gems with `bundle update` did _not_ change the test results. All passing but the one failure that works in real life outside of the test suite.

Tried tweaking the tests slightly to see if it's an issue with the test logic, but I've personally had no luck with that yet, either.

DeeDeeG:
If we can test this on a dyno on the main Heroku account for the project, that would add confidence in the Node 17+ workaround "just working" as I have it set in `app.json` without any fiddling with the Heroku dashboard.

If other maintainers can't help out with this, I'll be looking at paying $5 for an "eco dyno" on my personal Heroku account to see how it goes.

Would love to avoid paying Heroku money just for the privilege of testing the app, but that may just be how it is.

Will also let me run the Rails 7 upgrade PR through some paces on Heroku as well if I do that, though I'll note I haven't set up a postgresql DB in Heroku before, I don't think I'd know how, and that costs even more, so anything to do with actually hitting the DB won't work in my test dyno. That limits what I can really test with it, but at least it's something.

At minimum, I can confirm whether this workaround works. **Success criterion for testing: The app's JS compiles successfully.** (i.e. the app deploys to the dyno and doesn't go up in smoke, basically works.) The env var is only needed by Webpacker to compile, apparently. Or so I have heard. It may be possible to unset the env var after compiling, even. Anyway, this remains to be seen.

PR #691, opened 2024-04-20
Replace PhantomJS with Chromium

mmx900:
# Context

Digging the test failure of develop branch, I found that the project is still using old [PhantomJS](https://github.com/ariya/phantomjs)(with [Poltergeist](https://github.com/teampoltergeist/poltergeist)) which is outedated and doesn't support ES6 features. When I try to debug using `page.driver.debug`, I can see console errors including `Unexpected token 'const'`. As I remember there are many weird bugs in PhantomJS. So I just replaced that with Chromium/[Cuprite](https://github.com/rubycdp/cuprite)/[Ferrum](https://github.com/rubycdp/ferrum). Now all tests are passed in my local environment with `docker compose run -e "RAILS_ENV=test" web rake db:test:prepare spec`. But I'm not sure why it was OK before https://github.com/RefugeRestrooms/refugerestrooms/commit/2c70ab90249274eab8ee80366b3deb693eeb8907.

# Summary of Changes

- Remove outdated PhantomJS/Poltergeist dependency.
- Add Chromium to Dockerfile. (I considered Chrome, but the package was not available on Arm environment like linux in docker on apple silicon)
- Add Cuprite dependency with [no-sandbox option for docker.](https://github.com/rubycdp/cuprite?tab=readme-ov-file#install)
- Rewrite functions mocking location to resolve `Refuge is not defined` error.

# Checklist

- [ ] Tested Mobile Responsiveness
- [ ] Added Unit Tests
- [ ] CI Passes
- [ ] Deploys to Heroku on test Correctly (Maintainers will handle)
- [ ] Added Documentation (Service and Code when required)

PR #693, opened 2024-05-12
Add missing expectations to restrooms_spec

mmx900:
# Context
- Completed the unfinished tests and removed the comments that disabled RuboCop checks in restroom_spec, as mentioned on https://github.com/RefugeRestrooms/refugerestrooms/pull/691#discussion_r1590504890).

# Summary of Changes

- Added Oakland's information as a stub to Geocoder to resolve `unknown stub request` error in [restrooms_controller L30](https://github.com/RefugeRestrooms/refugerestrooms/blob/994fe4877b186eeecfc26d78001c6bbc1d6fcb09/app/controllers/restrooms_controller.rb#L30), which occured during the "It guesses my location" spec. The information is the same as declared in [guess_in_oakland.json](https://github.com/RefugeRestrooms/refugerestrooms/blob/994fe4877b186eeecfc26d78001c6bbc1d6fcb09/spec/fixtures/guess_in_oakland.json#L46-L51).
- Added a title to the current location marker to enable its selection during tests. But [as google deprecated google.maps.Marker](https://developers.google.com/maps/documentation/javascript/advanced-markers/migration?hl=en), I think that we should migrate it to use advanced marker API soon.
- Added a coordinate to `:oakland_restroom` factory to display the restroom when the map is focused on Oakland. Although there is already `:geocoded` trait, considering the name of fixture, I thought it was reasonable to add the coodinate.

# Checklist

- [ ] Added Unit Tests
- [ ] CI Passes

# Screenshots

### DIFF
--- a/spec/features/restrooms_spec.rb
+++ b/spec/features/restrooms_spec.rb
@@ -70,26 +70,22 @@ describe 'restrooms', :js do
       expect(find('button.current-location-button')['aria-label']).to be_truthy
     end
 
-    # rubocop:disable RSpec/NoExpectationExample
     it 'displays a map' do
       create(:oakland_restroom)
 
       visit root_path
       mock_location "Oakland"
       find('.current-location-button').click
-      # TODO: Figure out why this isn't working.
-      # print page.html
-      page.has_css?(".mapToggle", visible: true)
-      # find('.mapToggle').click
-      # print page.html
+      expect(page).to have_css(".map-toggle-btn", visible: :visible)
 
-      # TODO: Figure Out why This isn't working either
-      # page.has_css?(#)
-      # expect(page).to have_css('#mapArea.loaded')
-      # expect(page).to have_css('#mapArea .numberCircleText')
+      find('.map-toggle-btn').click
+
+      expect(page).to have_css('#mapArea.loaded')
+      expect(page).to have_css("#mapArea [role=button][aria-label='1']")
     end
   end
 
+  # rubocop:disable RSpec/NoExpectationExample
   describe 'preview' do
     it 'can preview a restroom before submitting' do
       visit "/"
--- a/spec/features/restrooms_spec.rb
+++ b/spec/features/restrooms_spec.rb
@@ -85,7 +85,6 @@ describe 'restrooms', :js do
     end
   end
 
-  # rubocop:disable RSpec/NoExpectationExample
   describe 'preview' do
     it 'can preview a restroom before submitting' do
       visit "/"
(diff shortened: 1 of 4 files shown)


spec/features/restrooms_spec.rb
```diff
@@ -59,7 +60,7 @@
       mock_location "Oakland"
       click_button 'Search by Current Location'
 
-      expect(page).not_to have_content 'Some Cafe'
+      expect(page).to have_content 'Some Cafe'
```
DeeDeeG, line 63: I suppose this `expect()` may have originally (before this Pull request) been written "wrong" (with inverted boolean value), so that it would pass originally, given the location aspects not being set up properly as the test scenario was originally written...

That said, given how strange the commit history of this test is, it would be reassuring and add to my confidence this is working correctly, if there was another restroom added for San Francisco, which we expect _not_ to show for a search in Oakland.

Something like this:

```rb
      create(:restroom, name: "Some San Francisco Bookshop")
```

...

```rb
      expect(page).not_to have_content 'Some San Francisco Bookshop'
```

Assuming that would work. Not strictly a blocker, and I can try to add this myself outside of the current Pull Request if that would be preferable. If I don't hear back I'll try and implement it myself when I get a moment to fire up Docker.

spec/features/restrooms_spec.rb
```diff
@@ -69,23 +70,18 @@
...
-      # print page.html
-      page.has_css?(".mapToggle", visible: true)
-      # find('.mapToggle').click
-      # print page.html
-
-      # TODO: Figure Out why This isn't working either
-      # page.has_css?(#)
-      # expect(page).to have_css('#mapArea.loaded')
-      # expect(page).to have_css('#mapArea .numberCircleText')
+      expect(page).to have_css(".map-toggle-btn", visible: :visible)
```
DeeDeeG, line 79: Just documenting something here in case anyone is confused like I was.

I was a little unclear about the difference between `true` and `:visible` here, but it turns out they are equivalent, per the docs:

https://rubydoc.info/github/teamcapybara/capybara/Capybara/Node/Finders#all-instance_method

> visible (Boolean, Symbol) — Only find elements with the specified visibility. Defaults to behavior indicated by [ignore_hidden_elements](https://rubydoc.info/github/teamcapybara/capybara/Capybara#configure-class_method).
> [ . . . ]
> - true - only finds visible elements.

> [ . . . ]
> - :visible - same as true; only finds visible elements.

DeeDeeG: Not sure why the CSS class for this button changed since the commented out test was written as `.mapToggle`, but indeed it did change to `.map-toggle-btn`! Thank you for fixing this.

DeeDeeG (approved):
Looks good to me, thank you for making these tests work!

PR #694, opened 2024-05-15
Fix rails-ujs import syntax causing thumbs up/down buttons to not work

mmx900:
# Context

- Currently there is `__webpack_require__(...).start is not a function` error in the JavaScript console. This error seems to have appeared after a recent upgrade to Rails 7. If I downgrade rails-ujs to 6.x, the error disappears. Additionally, it seems to break some functionallity that uses rails-ujs. This PR fixes the error and adds tests to prevent regression, as there were no existing tests for this functionality.

# Summary of Changes

- Fix js error by following [current documentation of rails-ujs](https://github.com/rails/rails/tree/7-1-stable/actionview/app/javascript#es2015).
- Add missing specs to prevent regression and ensure the voting functionality works as expected.

# Checklist

- [x] Added Unit Tests
- [x] CI Passes

# Screenshots

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -1,5 +1,7 @@
 // Vendor
-require('@rails/ujs').start()
+import Rails from '@rails/ujs'
+Rails.start()
+
 require('jquery')
 import 'bootstrap/dist/js/bootstrap'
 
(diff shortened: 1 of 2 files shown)


DeeDeeG (approved):
Thanks for this!

The syntax of the new `import` statement does indeed match the linked docs for rails-ujs.

New tests are much appreciated!

Tests (including new ones) are passing in CI. Looks good.

Thank you!

Merging this now.

